PR #13, opened 2021-02-25
Expand docs

tkkuehn:
Here's a first draft of some expanded documentation, filling out the overviews of the `bids` function and BIDS Apps, and adding pages on the config file and snakebids additions to Snakefile rule definitions.

It could still use a double check for accuracy and some editing for clarity, but it's a start.


akhanf (approved):
Looks great!

PR #17, opened 2021-03-22
upgrade urllib3 for security vulnerability

akhanf:


akhanf:
urllib removed from requirements in #14 so no longer an issue..

PR #18, opened 2021-03-23
Update docs to render entire API with fixed reST

tkkuehn:
This PR adds `app` and `snakemake_io` to the API documentation, and fixes some of the reST in some function docstrings that rendered in a wonky way, so the API docs now all render properly.


akhanf (approved):
awesome, thanks for this!

PR #19, opened 2021-03-24
Get readthedocs to install this package for autodoc

tkkuehn:
I think autodoc is failing on readthedocs because the Snakebids' external dependencies aren't being installed -- I think this small configuration change should resolve that (I don't know if there's a good way to test it without pulling the change first, though).


akhanf:
Thanks, I'll merge this in, but I think there is a way to build from a branch on readthedocs too (I've used that before, but I recall it was a hard option to find on the website!)

PR #20, opened 2021-03-24
Reformat to satisfy trivial requirements of pylint and flake8

tkkuehn:
I ran the the Python files in the project through pylint and flake8 as the start of a refactor to identify any style issues and bring the code in line with PEP 8. This PR is almost entirely adjusting small formatting issues or making minor one-line adjustments to reflect best practices (things like changing `var == None` to `var is None`). There are a few places remaining where pylint wants function complexity to be reduced, but I'll address those in a separate PR.

On a related note, it might be worthwhile to adopt a tool like [black](https://github.com/psf/black) or [autopep8](https://github.com/hhatto/autopep8) to automatically apply these kinds of formatting changes. An example of a GitHub action that automatically formats incoming PRs is [here](https://github.com/psf/Black/actions/runs/17913292/workflow).

### DIFF
--- a/snakebids/__init__.py
+++ b/snakebids/__init__.py
@@ -1,20 +1,37 @@
-from bids import BIDSLayout, BIDSLayoutIndexer
-import bids
+"""Utilities for converting Snakemake apps to BIDS apps."""
+
 import os
+from os.path import join
+from collections import OrderedDict
 import json
 import re
-import yaml
-import shutil
+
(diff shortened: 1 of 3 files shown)


akhanf:
Thanks, I agree we should definitely add that as an action! Feel free to add that in whenever you have time. We can also add similar actions to the cookiecutter app, though in that case having a snakemake linter would make more sense --

akhanf (approved):

PR #23, opened 2021-03-25
Run black -l 79 on project, and add GH action to lint with black

tkkuehn:
Adopting a code formatter ensures that code style is consistent across a project. Here, I've used [Black](https://github.com/psf/black) with the argument `-l 79` to bring all the Snakebids code to a style consistent with PEP 8.

This PR also adds a GitHub action that lints incoming PRs with black, to ensure that the incoming code is also formatted according to that style.

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -1,0 +1,13 @@
+name: Lint
+
+on: [push, pull_request]
+
+jobs:
+  lint:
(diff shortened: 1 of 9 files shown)


akhanf (approved):
Looks good, will have to get used to looking at " " instead of ' ' but I can live with that!

PR #24, opened 2021-03-25
Reduce complexity of top-level script

tkkuehn:
Pylint complained about the complexity of a few of the functions in `__init__.py`. This PR resolves those complaints, mostly by splitting out some functionality into smaller helper functions.


akhanf (approved):
this is great, thanks!!

PR #25, opened 2021-03-25
Suffix wildcards

akhanf:
- adds ability to use suffix as a wildcard
- exposes wildcards to CLI with --wildcards_<input_name> options (like --filter_*)

### DIFF
--- a/snakebids/__init__.py
+++ b/snakebids/__init__.py
@@ -720,6 +720,7 @@ def __get_lists_from_bids(
                         # bids_tags.json, where e.g. acquisition -> acq is
                         # defined.. -- then, can use wildcard_name instead
                         # of out_name..
+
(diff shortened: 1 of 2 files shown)


akhanf:
@tkkuehn had a go at this after trying out  snakebids2dicom.

The edit needed was actually fairly minimal, just needed to deal with suffix as a special case and regex differently, but while I was at it I exposed the wildcards list to the CLI too (will be more generally useful).. 

I've also edited snakebids2dicom with this change, look out for a PR soon on that, and will be pushing some data to orthanc server too

tkkuehn:
Looks great! Would be good to add a test to cover this case, but otherwise I think this is ready to merge.

akhanf:
Thanks for adding the test @tkkuehn, I'll merge this in now!

PR #36, opened 2021-04-01
removed snakemake usage from help

akhanf:
added --help_snakemake instead, to simplify the default usage and to fix the issue where [targets] from snakemake would appear before the required snakebids arguments in the combined usage.

Note that we now lose the functionality of checking whether any
namespace conflicts exist between snakebids and snakemake arguments, so will be up to the workflow author to ensure that is the case.


akhanf:
Note I've kept the combined parser as an attribute, since it may still be useful for generating a boutiques descriptor down the road..

tkkuehn (approved):
Looks good to me, and worked on my machine!

PR #38, opened 2021-04-07
Use exceptions and logging to handle errors

tkkuehn:
This PR adds exceptions to handle errors, replacing the previous behaviour of exiting the program, and allowing calling functions to handle them if desired.

This also adopts the standard `logging` module, which allows a user to configure which logging messages are shown in a standard way.


akhanf (approved):
This is great!

PR #47, opened 2021-05-13
filters by participant_label when pybids skipped

akhanf:
note: still need to filter input lists with exclude_participant_label in
this case, and may need to also filter when a mix of pybids and --path_*
are used..


akhanf:
closes #47

PR #49, opened 2021-06-15
better configfile support (+json), yaml formatting

akhanf:
- uses snakemakes configfile loader for json + yaml support
- wider range of configfile names supported, input path used for output
instead of hard-coded as config/snakebids.yml
- better yaml writing (no resorted keys, and no extra python ordereddict
object strings cluttering the output)..

TODO: 
- [ ] ~~configfile history (was thinking of writing a time-stamped
config, perhaps to a hidden directory (.snakebids_history?).. time hash
function created but unused as of yet..~~
- [x] testing on existing workflows to make sure we still have backwards
compatibility
- [x] merge in tutorial branch into here (since that motivated this change)
- [ ] ~~add some more tests?~~
- [ ] ~~update documentation~~

(will just create issues for those instead so can release before ohbm hackathon)


akhanf:
works for hippunfold

PR #52, opened 2021-09-01
Resolve arguments annotated with type: Path 

pvandyken:
# Problem
Currently, when supplying arguments to the command line executable (e.g. `run.py`), the `input` and `output` positional arguments are resolved (e.g. `./data` becomes `/home/user/folder/data`) and a new `config.yaml` is generated in the output folder. All keyword arguments (e.g. `--keyword argument`) are printed into the new `config.yaml` as `keyword: argument`. Snakemake is then run using the new config file as config and using the output dir as the working directory. The equivalent snakemake call would be 
```
snakemake --configfile /path/to/output/dir/config/config.yaml --directory /path/to/output/dir
```

Because the working directory is switched, any additional relative paths supplied to the original `run.py` call will now be interpreted relative to the output directory. For instance, if the user had an argument `--supporting-data ./path/to/data`, this path would be evaluated relative to the output folder. Snakemake would search for the data at `/path/to/output/dir/path/to/data`. The user, however, would expect snakemake to look for the folder at a location relative to where `run.py` was called.

# Solution
Argparse allows arguments to be annotated as a specific type such as `str`, `int`, etc. The solution here was to allow users to annotate arguments with either `Path` or `pathlib.Path`. `app.py` will convert these files to `Path`s and resolve them before printing them into the new `config.yaml`. Note that any arguments not annotated with Path will not be affected, even if they look like paths. 

If the solution is green-lit, I can update the documentation to reflect this new behaviour. EDIT: I added some documentation

A final side-note, I refactored `app.py` to use the more modern `pathlib` module rather than `os.path`. Should have been its own issue, but it's baked into the code now. Everything's been tested and checks out.

### DIFF
--- a/snakebids/app.py
+++ b/snakebids/app.py
@@ -236,6 +236,14 @@ class SnakeBidsApp:
 
         # update the parser with config options
         for name, parse_args in self.config["parse_args"].items():
+            if "type" in parse_args:
+                type_ = parse_args["type"]
+                if not type_ in globals():
+                    raise TypeError(
+                        f"{type_} is not available "
+                        f"as a type for {name}"
+                    )
+                parse_args["type"] = globals()[parse_args["type"]]
             app_group.add_argument(name, **parse_args)
 
         # general parser for
--- a/snakebids/app.py
+++ b/snakebids/app.py
@@ -3,7 +3,6 @@
 """Tools to generate a Snakemake-based BIDS app."""
 
 import os
-from pathlib import Path
 import pathlib
 import subprocess
 import argparse
@@ -13,6 +12,8 @@ import yaml
 import bids
 import snakemake
 from snakemake.io import load_configfile
+
+Path = pathlib.Path
 
 bids.config.set_option("extension_initial_dot", True)
 logger = logging.Logger(__name__)
--- a/snakebids/app.py
+++ b/snakebids/app.py
@@ -92,13 +92,28 @@ def get_time_hash():
     return hash.hexdigest()[:8]
 
 def resolve_path(path_candidate):
+    """Helper function to resolve any paths or list
+    of paths it's passed. Otherwise, returns the argument 
+    unchanged. 
+
+    Parameters
+    ----------
+    command : list, os.Pathlike, object
+        command to run
+
+    Returns
+    -------
+    list, os.Pathlike, object
+        If os.Pathlike or list  of os.Pathlike, the same paths resolved.
+        Otherwise, the argument unchanged.
+    """
     if isinstance(path_candidate, list):
         return [resolve_path(p) for p in path_candidate]
-    elif isinstance(path_candidate, os.PathLike):
-        
+
+    if isinstance(path_candidate, os.PathLike):
         return path_candidate.resolve()
-    else:
-        return path_candidate
+    
+    return path_candidate
 
 SNAKEFILE_CHOICES = [
     "Snakefile",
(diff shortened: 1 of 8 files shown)


akhanf:
Thanks Peter! I'll try to take a look at this sometime soon, and I've requested Tristan to do the same when he gets a chance.

snakebids/app.py
```diff
@@ -3,15 +3,18 @@
...
 import os
+import pathlib
 import subprocess
 import argparse
 import logging
 import sys
 import yaml
 import bids
-from snakemake import get_argument_parser
+import snakemake
```
akhanf, line 13: any good reason to import the entire snakemake module rather than just the `get_argument_parser` function?

akhanf (commented):
This all looks great to me, but I will let @tkkuehn have a closer look for formatting and the test code.

snakebids/app.py
pvandyken: That was solely for stylist reasons: it makes it more clear where get_argument_parser comes from. I'm not sure if it introduces a performance cost. If it does, or you just prefer it the old way, it can definitely be reverted.

pvandyken:
I forgot that my new test code requires pytest-mock. I added it as a dependency into the git workflow.

snakebids/app.py
```diff
@@ -88,7 +91,14 @@ def get_time_hash():
     hash.update(str(time.time()).encode('utf-8'))
     return hash.hexdigest()[:8]
 
-
+def resolve_path(path_candidate):
+    if isinstance(path_candidate, list):
+        return [resolve_path(p) for p in path_candidate]
+    elif isinstance(path_candidate, os.PathLike):
```
tkkuehn, line 97: This can just be an `if` (because the previous `if` statements returns if it triggers.

snakebids/app.py
```diff
@@ -88,7 +91,14 @@ def get_time_hash():
...
+        
+        return path_candidate.resolve()
+    else:
```
tkkuehn, line 100: This `else` can be removed.

snakebids/app.py
```diff
@@ -240,6 +245,15 @@ def __create_parser(self, include_snakemake=False):
 
         # update the parser with config options
         for name, parse_args in self.config["parse_args"].items():
+            # Convert type annotations froms strings to class types
+            if "type" in parse_args and isinstance(parse_args["type"], str):
```
tkkuehn, line 249: I think this would be slightly more readable as something like:

```python
if "type" in parse_args:
    try:
        parse_args["type"] = globals()[parse_args["type"]]
    except KeyError as err:
        raise TypeError(
            f"{parse_args['type']} is not available "
            + f"as a type for {name}"
        ) from err
```

snakebids/app.py
tkkuehn: I don't think there would be a performance cost -- the old way makes it clear what we're using from snakemake, but either way is fine by me.

tkkuehn (changes requested):
Thanks for the pull request! This fixes an ongoing problem nicely, and I appreciate the included tests.

I have a couple of nitpicky comments but this largely looks good to me.

snakebids/app.py
```diff
@@ -3,15 +3,18 @@
...
 import argparse
 import logging
 import sys
 import yaml
 import bids
-from snakemake import get_argument_parser
+import snakemake
 from snakemake.io import load_configfile
 
+Path = pathlib.Path
```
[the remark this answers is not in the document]
pvandyken: I typically do the same. My reasoning here was that users could define type in the config as either "Type" or "pathlib.Type". However, this adds very little convenience and probably wouldn't even make it into documentation regardless, so we can definitely do it as you wrote

pvandyken:
> I don't think there would be a performance cost -- the old way makes it clear what we're using from snakemake, but either way is fine by me.

That's a fair point. I'll let you make the final call on that one

snakebids/app.py
pvandyken: I'll change it to something like that. I'd still like to check that `"type"` is a `str`, though: I've hit edge cases where `parse_args["type"]` has already been resolved to the class it represents on some previous iteration, but then fails because it's not a string.

tkkuehn: Sounds good!

tkkuehn:
> I'll let you make the final call on that one

You can leave it as it is now!

snakebids/app.py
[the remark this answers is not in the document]
tkkuehn: Ah, I see what you mean -- This way, "Path" and "pathlib,Path" are in `globals()`. With that explanation, this makes sense to me!

akhanf:
thanks both, looks all good to me now, so will merge this in

PR #55, opened 2021-09-14
Accepts pathlib.Path as root in bids function

pvandyken:
Previously, `bids()` only accepted `str`. This extends it to allow `pathlib.Path`s.

I also refactored the `os.path.join()` calls `__init__.py` to use pathlib.


akhanf (commented):
looks good to me, thanks!

PR #58, opened 2021-10-08
Fixes for pybids dataset_description.json compatibility

akhanf:
Pybids is still a bit behind when it comes to parsing BIDS derivatives datasets, since the new BIDS spec (that e.g. fmriprep uses) uses a `GeneratedBy` list in the dataset_description.json, which supercedes the older `PIpelineDescription.Name` used by Pybids to parse a dataset. 

This PR adds a tool to add in a PipelineDescription to an existing dataset_description file, using the first entry of the GeneratedBy list to get the name and version. This is so we can more easily use BIDS derivatives in snakebids. Once Pybids is fixed to deal with this properly, we can ditch this, but it solves the problem for now. 

I've also added the PIpelineDescription field to our template project files, so that nakebids outputs are also compatible with pybids/snakebids when parsing as a derivative.

### DIFF
--- a/snakebids/fix_dd.py
+++ b/snakebids/fix_dd.py
@@ -9,7 +9,7 @@ def get_parser():
         description="Update BIDS Derivatives Dataset  to be backwards compatible with pybids",
         epilog=("Adds PipelineDescription.Name to the dataset_description.json file, from earlier BIDS spec) that pybids requires, using the first entry of the GeneratedBy list.  This is needed if parsing a modern version of fmriprep derivatives with snakebids/pybids."),
     )
-    parser.add_argument("dataset_root", help="BIDS derivative dataset to fix")
+    parser.add_argument("dataset_root","dataset-root", help="BIDS derivative dataset to fix")
     return parser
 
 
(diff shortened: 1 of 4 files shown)


snakebids/fix_dd.py
```diff
@@ -0,0 +1,53 @@
...
+import json
+import argparse
+
+
+def get_parser():
+    parser = argparse.ArgumentParser(
+        description="Update BIDS Derivatives Dataset  to be backwards compatible with pybids",
+        epilog=("Adds PipelineDescription.Name to the dataset_description.json file, from earlier BIDS spec) that pybids requires, using the first entry of the GeneratedBy list.  This is needed if parsing a modern version of fmriprep derivatives with snakebids/pybids."),
+    )
+    parser.add_argument("dataset_root", help="BIDS derivative dataset to fix")
```
pvandyken, line 12: Not sure what the style has been in other parts of the app, but it's nice to make `dataset-root` an alias

tkkuehn (approved):
Looks good to me!

snakebids/fix_dd.py
akhanf: added

akhanf:
merging this in, will have to address creating the dataset_description.json in another PR

PR #60, opened 2021-11-01
Use poetry for dependency management

pvandyken:
This would replace setup.py with a pyproject and use Poetry as a more modern and sustainable approach to dependency management.

Development and test dependencies previously not explicitly required are now listed.

Github workflow is updated to use poetry for publishing and testing. The workflows are based on previous work I did, so they *should* work, but that will need to be tested. I used `secrets.PYPI_API_TOKEN` as on the previous workflow with twine. I assume it's a pypi api key scoped to the snakebids project?


akhanf (approved):
Thanks, this looks great! 
Look forward to getting a quick demo of setting up poetry on Friday (eg how the .lock file was created).

akhanf:
Yes, that token is a github secret for this repository, so the CI should work still work -- will see!

pvandyken:
Once we're ready to merge, I think this should go under 0.4.0-dev. I'm currently wrapping up the basic code for changing output modes (watch for a PR later this week), and once that and everything related is ready, we can push a full 0.4 release?

pyproject.toml
```diff
@@ -0,0 +1,34 @@
...
+pybids = ">=0.13.2"
+snakemake = ">=5.28.0"
+PyYAML = ">=5.3.1"
+cookiecutter = ">=1.7.2"
+
+[tool.poetry.dev-dependencies]
+black = "^21.10b0"
+pytest = "^6.2.5"
+ConfigArgParse = "^1.5.3"
+pytest-mock = "^3.6.1"
```
tkkuehn, line 27: Can we add `pylint = "^2.11.1"` here? I like to use it locally and I don't think poetry has a way to install it separately,

tkkuehn (changes requested):
Looks good to me! One small request if there are no objections, but otherwise this looks good to merge.

pyproject.toml
kaitj: Additionally, for packages using inequality requirements, I think I would lean towards using either caret requirements or multiple requirements so that it doesn't automatically install a different major version that is more likely to introduce changes that aren't backwards compatible (if version in the `.lock` file or the `.lock` file itself isn't found for some reason).

pvandyken: Sure thing. Actually, it would be good to agree on what code style/quality/linters we'd like to use in the project here-on in + set up appropriate checks in gh-actions. I'll open up a new issue to discuss

pvandyken:
If it's alright @tkkuehn and @kaitj, can I go ahead and merge this and we'll address the points in a new PR once we get #61 worked out (just because it depends on this commit)

PR #61, opened 2021-11-05
Implement workflow and bidsapp mode

pvandyken:
## TODOs
- [x] Update sphinx documentation
- [x] Improve exception handling (e.g. by catching `RunError` and printing a nicely formatted message without the trace)
- [x] Finalize handling of the `.snakemake` folder.
- [x] Prevent deletion of `.snakemake` folder when converting from workflow to bidsapp mode
- [x] Bump version to 0.4.0a1 and disable bumpversion for now

## Overview
With this PR, snakebids will, by default, produce output in bidsapp mode, meaning snakemake results will be placed directly in the output folder. The user, via command line flag `--workflow-mode` or `-w`, can change the output to "workflow mode" which will cause the entire snakebids app to be copied to the output dir. Snakemake results will be put into the `output/results` folder.

## Problem
Currently, snakebids produces output in roughly a bidsapp fashion, but still produces a `config/config.yml` file in the output directory. There is no convenient way to use the `run.py` function in a manner analogous to snakemake, that is, to run a pipeline progressively and iteratively, collecting the outputs in the results folder. At the same time, it doesn't particularly conform to bidsapp specs either, with the production of an extra config folder in the output.

### bidsapp and workflow modes
bidsapp mode is intended to emulate a typical bids app. The application produces just the ouptut, placed in the folder chosen by the user. Because it's the most transparent mode, and doesn't change the users output path or add extra files, it's set to be the default. The user must opt into workflow mode using a CLI flag. To keep a record of runtime parameters, the config file is saved into the `output/code` folder, in accordance with the bids spec. Successive runs of the snakebids app will overwrite the config file saved into the code folder unless the name of the config file is changed by the user.

Workflow mode is intended for modification and development of the pipeline. By setting the CLI flag, the snakebids app (parent) will reproduce itself (a child) in the output directory specified by the user, putting the snakemake results in the `output/results` folder. This newly created copy of the snakebids app will be set as the working directory for snakemake. This mode is not intended to be run multiple times in succession. Rather, Snakemake should be run directly in the child app. If the user runs the parent app in workflow mode a second time, no files will be copied over except for the `config.yaml` file being used for the run. However, if the parent finds a `config.yaml` of the same name already in the child, it will throw an error.

Both in bidsapp and in workflow modes, the appropriate directory for results will be made available to the Snakefile in the variable: `config["bids_dir"]`.

### .snakebids file
To track the mode of output directories, snakebids generates a `.snakebids` file at the top level of each output. This is a simple json file that currently has just one field: mode. In the future, this file should be expanded to include more metadata and tracking.

Using the `.snakebids` file, snakebids can convert between bidsapp and workflow mode, and vice versa. Note that in the conversion from workflow to bidsapp mode, all of the workflow files, metadata, and everything that's not in the results folder in the child app will be deleted. As such, snakebids will refuse to perform such a conversion unless provided with the `--force-conversion` flag.

As extra precautions, snakebids will refuse to make output in a directory already containing items. It will not convert between bidsapp and workflow modes (or vice versa) if the snakemake results itself contains a results folder. 

### Running in itself
If the parent app itself is set as the output directory, Snakebids will automatically switch to workflow mode, alerting the user of this fact if they did not specify such. Results will, as usual, be put in the results folder.

### Retrofit function
To facilitate the conversion of legacy outputs to the new format, snakebids provides a --retrofit flag.
Running this with an appropriate output, Snakemake will delete any config files found in the output and generate a new `.snakebids` file. The user will be prompted with the list of files to be deleted for confirmation. This will not work if unrecognized files (anything not a config file in the list in app.py) are found in the `output/config` folder.

Retrofit puts the output in bidsapp mode, but the user can switch to workflow mode immediately just by adding the `-W` flag.

## Limitations
- Because of the current simplicity of the `.snakebids` file, Snakebids cannot currently distinguish between outputs generated by the current app or different apps. This can be improved in future editions.

## Omnibus Issues
These technically should have got their own PRs, but got merged in along the way:
- Private functions were defined using double underscores: `__method()`. This has a special meaning in Python and causes mangling of the method name, making it difficult to test ([see here](https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-single-and-double-underscore-before-an-object-name)). These were thus switched to the more canonical single underscore `_method()`. 
- To make way for a new exception and prevent confusion, all exceptions were moved to their own module
- The import syntax for `pybids.config` was changed to satisfy my IDE linter.  `pybids.config` is not directly importable, technically (although it seemed to work anyhow?). At any rate, the current syntax is cleaner and gives me no squiggly lines.
- Snakemake is now run directly, rather than using a subprocess. This allows colors to still be displayed.

### DIFF
--- a/snakebids/app.py
+++ b/snakebids/app.py
@@ -481,6 +481,7 @@ class SnakeBidsApp:
         else:
             new_config_file = self.outputdir/"code"/Path(self.configfile_path).name
             cwd = self.snakemake_dir
+            force_config_overwrite = True
 
         write_config_file(new_config_file, self.config, force_config_overwrite)
 
--- a/snakebids/app.py
+++ b/snakebids/app.py
@@ -1,22 +1,18 @@
 """Tools to generate a Snakemake-based BIDS app."""
 
-import json
 import os
 import pathlib
 import subprocess
 import argparse
 import logging
 import sys
-import yaml
-import itertools as it
-import shutil
 
 import snakemake
 from snakemake.io import load_configfile
 from colorama import Fore
 
 from bids import config as bidsconfig
-from snakebids.exceptions import ConfigError
+from snakebids.exceptions import ConfigError, RunError
 from snakebids.output import prepare_output, retrofit_output, write_config_file, write_output_mode
 
 # We define Path here in addition to pathlib to put both variables in globals()
@@ -259,19 +255,34 @@ class SnakeBidsApp:
         parser.add_argument(
             "--workflow-mode",
             "-W",
-            action="store_true"
+            action="store_true",
+            help=(
+                "Run Snakebids in Workflow mode. This will cause the entire Snakebids "
+                "app, except for the results folder, to be copied into your "
+                "output_dir. Snakemake will be run in this new child app, and results "
+                "will be put in output_dir/results."
+            )
         )
 
         # We use -x as the alias because both -f and -F are taken by snakemake
         parser.add_argument(
             "--force-conversion",
             "-x",
-            action="store_true"
+            action="store_true",
+            help=(
+                "Force snakebids to convert a workflow output to a bidsapp output. "
+                "conversion will delete all the files in the workflow snakebids app."
+            )
         )
 
         parser.add_argument(
             "--retrofit",
-            action="store_true"
+            action="store_true",
+            help=(
+                "Convert a legacy Snakebids output (Snakebids 0.3.x or lower) into "
+                "bidsapp mode. This will delete any config files currently in the "
+                "output."
+            )
         )
 
         # add option for printing out snakemake usage
@@ -446,8 +457,8 @@ class SnakeBidsApp:
                     f"{Fore.YELLOW}You specified your output to be in the snakebids "
                     "directory, so we're switching automatically to workflow mode!\n"
 
-                    f"{Fore.RESET}You'll find your results in the "
-                    "`snakebidsdir/results` directory."
+                    f"{Fore.RESET}You'll find your results in "
+                    f"`{(self.snakemake_dir/'results').resolve()}`"
                 )
         else:
             mode = "workflow" if self.workflow_mode else "bidsapp"
@@ -464,14 +475,19 @@ class SnakeBidsApp:
                         for p in CONFIGFILE_CHOICES if (self.outputdir/p).exists()
                     )
                 ):
+                    print(f"{Fore.YELLOW}Exiting. No conversion performed.{Fore.RESET}")
                     exit(1)
         
-        root = prepare_output(
-            self.snakemake_dir,
-            self.outputdir,
-            mode,
-            self.force
-        )
+        try:
+            root = prepare_output(
+                self.snakemake_dir,
+                self.outputdir,
+                mode,
+                self.force
+            )
+        except RunError as err:
+            print(err.msg)
+            exit(1)
 
         if mode == "workflow":
             self.config["output_dir"] = str(root)
--- a/snakebids/app.py
+++ b/snakebids/app.py
@@ -13,6 +13,8 @@ import bids
 import snakemake
 from snakemake.io import load_configfile
 
+from snakebids.exceptions import ConfigError
+
 # We define Path here in addition to pathlib to put both variables in globals()
 # This way, users specifying a path type in their config.yaml can indicate
 # either Path or pathlib.Path
@@ -22,12 +24,6 @@ bids.config.set_option("extension_initial_dot", True)
 logger = logging.Logger(__name__)
 
 
-class ConfigError(Exception):
-    """Exception raised for errors with the Snakebids config."""
-
-    def __init__(self, msg):
-        self.msg = msg
-        Exception.__init__()
 
 
 class KeyValue(argparse.Action):
--- a/snakebids/app.py
+++ b/snakebids/app.py
@@ -214,16 +214,16 @@ class SnakeBidsApp:
         self.config["snakemake_dir"] = snakemake_dir
         self.config["snakefile"] = self.snakefile
 
-        self.parser_include_snakemake = self.__create_parser(
+        self.parser_include_snakemake = self._create_parser(
             include_snakemake=True
         )
-        self.parser = self.__create_parser()
+        self.parser = self._create_parser()
 
         if not skip_parse_args:
-            self.__parse_args()
+            self._parse_args()
 
 
-    def __create_parser(self, include_snakemake=False):
+    def _create_parser(self, include_snakemake=False):
         """Create a parser with snakemake parser as parent solely for
         displaying help and checking conflicts, but then for actual parsing
         use snakebids parser to parse known args, then pass remaining to
@@ -341,7 +341,7 @@ class SnakeBidsApp:
 
         return parser
 
-    def __parse_args(self):
+    def _parse_args(self):
 
         # use snakebids parser to parse the known arguments
         # will pass the rest of args when running snakemake
--- a/snakebids/app.py
+++ b/snakebids/app.py
@@ -1,5 +1,6 @@
 """Tools to generate a Snakemake-based BIDS app."""
 
+import json
 import os
 import pathlib
 import subprocess
@@ -7,12 +8,16 @@ import argparse
 import logging
 import sys
 import yaml
-import bids
+import itertools as it
+import shutil
+
 import snakemake
 from snakemake.io import load_configfile
+from colorama import Fore
 
 from bids import config as bidsconfig
 from snakebids.exceptions import ConfigError
+from snakebids.output import prepare_output, retrofit_output, write_config_file, write_output_mode
 
 # We define Path here in addition to pathlib to put both variables in globals()
 # This way, users specifying a path type in their config.yaml can indicate
@@ -80,14 +85,7 @@ def run(command, env=None):
     if process.returncode != 0:
         raise Exception("Non zero return code: %d" % process.returncode)
 
-def get_time_hash():
-    """ currently unused """
-    import hashlib
-    import time
 
-    hash = hashlib.sha1()
-    hash.update(str(time.time()).encode('utf-8'))
-    return hash.hexdigest()[:8]
 
 def resolve_path(path_candidate):
     """Helper function to resolve any paths or list
@@ -122,14 +120,19 @@ SNAKEFILE_CHOICES = [
 
 
 CONFIGFILE_CHOICES = [
-            "config/snakebids.yml",
-            "config/snakebids.json",
-            "snakebids.yml",
-            "snakebids.json",
-            "config.yml",
-            "config.json",
-            "config/config.json",
-            "config/config.yml"]
+    "config/snakebids.yml",
+    "config/snakebids.yaml",
+    "config/snakebids.json",
+    "snakebids.yml",
+    "snakebids.yaml",
+    "snakebids.json",
+    "config.yml",
+    "config.yaml",
+    "config.json",
+    "config/config.json",
+    "config/config.yml",
+    "config/config.yaml"
+]
 
 
 class SnakeBidsApp:
@@ -165,9 +168,20 @@ class SnakeBidsApp:
     configfile_path : str
         Relative path to config file (relative to snakemake_dir)
     updated_config : str
-        Absolute path to the updated config file to write
-
+        Absolute path to the updated config file to write.
+    workflow_mode : bool
+        If true, produces output in workflow mode, otherwise snakemake mode.
+    force : bool
+        If false, prevents conversion of output from workflow mode to bidsapp mode.
+    outputdir : Path
+        Path to outputdir specified by user
+    retrofit : bool
+        If true, run_snakemake will attempt to convert legacy output format into bidsapp
+        format.
     """
+
+    outputdir: Path
+    snakemake_dir: Path
 
     def __init__(self, snakemake_dir, skip_parse_args=False):
         # input argument is the dir where snakemake would be run
@@ -176,13 +190,15 @@ class SnakeBidsApp:
         # generated_config, so the workflow can use it to source files from
         # it (e.g. atlases etc..)
 
+        self.snakemake_dir = Path(snakemake_dir).resolve()
+
         # look for snakebids.yml in the snakemake_dir, quit if not found
-        self.configfile_path = None
+        self.configfile_path = ""
         for path in CONFIGFILE_CHOICES:
-            if Path(snakemake_dir, path).exists():
+            if (self.snakemake_dir / path).exists():
                 self.configfile_path = path
                 break
-        if self.configfile_path is None:
+        if not self.configfile_path:
             raise ConfigError(
                 f"Error: no config file found, tried {', '.join(CONFIGFILE_CHOICES)}."
             )
@@ -190,7 +206,7 @@ class SnakeBidsApp:
         # look for snakefile in the snakemake_dir, quit if not found
         self.snakefile = None
         for snakefile_path in SNAKEFILE_CHOICES:
-            if Path(snakemake_dir, snakefile_path).exists():
+            if (self.snakemake_dir / snakefile_path).exists():
                 self.snakefile = Path(snakemake_dir, snakefile_path)
                 break
         if self.snakefile is None:
@@ -239,6 +255,24 @@ class SnakeBidsApp:
             parser = argparse.ArgumentParser(
                 description="Snakebids helps build BIDS Apps with Snakemake"
             )
+
+        parser.add_argument(
+            "--workflow-mode",
+            "-W",
+            action="store_true"
+        )
+
+        # We use -x as the alias because both -f and -F are taken by snakemake
+        parser.add_argument(
+            "--force-conversion",
+            "-x",
+            action="store_true"
+        )
+
+        parser.add_argument(
+            "--retrofit",
+            action="store_true"
+        )
 
         # add option for printing out snakemake usage
         parser.add_argument(
@@ -349,6 +383,11 @@ class SnakeBidsApp:
             k: resolve_path(v) for k, v in args.__dict__.items()
         }
 
+        self.workflow_mode = args.workflow_mode
+        self.force = args.force_conversion
+        self.outputdir = Path(args.output_dir).resolve()
+        self.retrofit = args.retrofit
+
         # add snakebids arguments to config
         self.config.update(args.__dict__)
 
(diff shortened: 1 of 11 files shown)


akhanf:
This looks super!
I'll try to give it a whirl myself at some point today

snakebids/app.py
```diff
@@ -1,33 +1,33 @@
-#!/usr/bin/env python3
-
 """Tools to generate a Snakemake-based BIDS app."""
 
+import json
```
tkkuehn, line 3: This doesn't seem to be used.

snakebids/app.py
```diff
@@ -1,33 +1,33 @@
...
 import os
 import pathlib
 import subprocess
 import argparse
 import logging
 import sys
 import yaml
-import bids
+import itertools as it
```
tkkuehn, line 11: This and `shutil` also don't seem to be used.

snakebids/app.py
```diff
@@ -245,6 +256,24 @@ def __create_parser(self, include_snakemake=False):
                 description="Snakebids helps build BIDS Apps with Snakemake"
             )
 
+        parser.add_argument(
```
tkkuehn, line 259: It would be good to add help documentation to these arguments.

snakebids/app.py
```diff
@@ -396,70 +430,76 @@ def __parse_args(self):
...
+            write_output_mode(self.outputdir/".snakebids", "workflow")
+            mode = "workflow"
+            force_config_overwrite = True
+            if self.workflow_mode == False:
+                print(
+                    f"{Fore.YELLOW}You specified your output to be in the snakebids "
+                    "directory, so we're switching automatically to workflow mode!\n"
+
+                    f"{Fore.RESET}You'll find your results in the "
+                    "`snakebidsdir/results` directory."
```
tkkuehn, line 450: Would it be possible to resolve this path to give the actual path here?

snakebids/app.py
```diff
@@ -396,70 +430,76 @@ def __parse_args(self):
...
+                        self.configfile_path = path
+                if not retrofit_output(
+                    self.outputdir,
+                    # Find all config files in the outputdir
+                    (
+                        self.outputdir/p 
+                        for p in CONFIGFILE_CHOICES if (self.outputdir/p).exists()
+                    )
+                ):
+                    exit(1)
```
tkkuehn, line 467: An extra print statement here logging that we're exiting here because the user refused to have leftover files deleted would be helpful, I think.

tkkuehn (commented):
This looks good at a high level but I didn't have time today to finish an in-depth review. Here are some initial comments, though! (Also, it would be good to update the sphinx documentation to cover the split between BIDS App and workflow mode).

pvandyken:
Thanks for the review @tkkuehn , I agree with all the points and will implement them as soon as possible.

pvandyken:
Also, apologies for the weird sentence formatting in the initial message, I missed that when transferring from my markdown editor. It's been fixed for posterity.

pvandyken:
One issue does need to be addressed: when retrofitting, this PR will not currently touch any `.snakemake` folder in the output directory. The rationale is that this may be the only record of the user's runs, and we don't want to delete that. However, this is not really satisfactory. By default, retrofitting puts the output in bidsapp mode, so future `.snakemake` folders will be generated in the parent app directory. Thus, the saved .snakemake folder won't even be used. On the other hand, if the app is switched into workflow mode and a `.snakemake` folder resides in the parent app, it will be copied to the child app, and the saved `.snakemake` folder will be moved to the `results` folder with the rest of the pre-existing files in the output (sorry, this is a little convoluted, happy to try describing again if this doesn't make sense). 

In either scenario, the reserved `.snakemake` folder is unused.

A solution to this dilemma may be to check both parent and output directories for these `.snakemake` folders, when retrofitting. If only one is found, fine and good and we can deal with it. If two are found, we force the user to choose one (or exit the app). That `.snakemake` folder will be held as authoritative.

Any thoughts on this @tkkuehn @akhanf?

akhanf:
Just trying things out now --

Some comments:

1. In bidsapp mode, I don't think it is right to have the`.snakemake` folder exist in the original workflow folder.

It unnecessarily ties runs of the same app, but on different datasets and different output folders. E.g. say I install a snakebids app run.py as a console script, let's say it is hippunfold. Then if I wanted to run hippunfold on two completely independent datasets:
```
hippunfold bids_datasetA bids_datasetA/derivatives/hippunfold participant 
hippunfold bids_datasetB bids_datasetB/derivatives/hippunfold participant 
```
These would share the same `.snakemake` folder (located wherever the workflow is installed), so would have issues with locking/unlocking the workflow, the `.snakemake` folder would unnecessarily know about files in both output folders, etc..

Also, depending on how he snakebids app was installed, the `.snakemake` folder could be either be tucked away in a virtualenv (not an obvious place for it), or if we have the entire app packaged in a single container (like hippunfold is), it would be in a read-only location (could perhaps be corrected with a bind path when running, but would be ad-hoc..).


So, I think it would be better to set the snakemake working directory to outputdir in bidsapp mode too. 


2. If we make that change, then in bidsapp mode there is an issue with resolving the configfile path in the snakemake workflow, e.g. `configfile: config/snakebids.yml` won't be found since relative to the working directory we would have `code/snakebids.yml`.  

A fix for this could be to use snakemake's (somewhat recently added?) ability to get paths relative to the Snakefile: 
```
configfile: workflow.source_path('../config/snakebids.yml')
```

3. Unrelated to the above, instead of having workflow authors use `bids=(root=config['output_dir'] ...`, which results in absolute file paths for all the workflow outputs, I think it would be better to have a new config variable that is either set to `''` (when in bidsapp mode) or `'results'` (when in workflow mode).. This retains the use of relative file paths for any workflow outputs, which is what snakemake recommends in general (for portability). I've gone ahead and made `config['root']`  for this, but open to suggestions on the naming.. 

I've made these changes in my commit, along with edits to test cases to accommodate the changes. I've also updated the project_template snakebids app to comply too. Note, we should perhaps add some test cases that spawn and use this template too.

snakebids/output.py
```diff
@@ -0,0 +1,472 @@
...
+
+    Returns
+    -------
+    Path
+        Path of results folder in new snakemake location
+
+    """
+    # shutils.copytree makes it hard to exclude just root level folders, so we do this
+    # manually
+    old_cwd = Path().cwd()
```
tkkuehn, line 310: Nitpick: Can just use `Path.cwd()`, don't need to instantiate the path.

snakebids/output.py
```diff
@@ -0,0 +1,472 @@
...
+    os.chdir(src)
```
tkkuehn, line 311: Do we need to change the working directory? I would generally prefer to avoid this kind of thing if possible and it seems to me like we could iterate over `Path(src).iterdir()`.

snakebids/output.py
```diff
@@ -0,0 +1,472 @@
...
+    
+    # Convert to workflow mode
+    if end == "workflow":
+        print(
+            f"{Fore.GREEN}Found bidsapp output at {Fore.RESET}{output.resolve()}\n\n"
+            f"{Fore.GREEN}Converting to workflow mode...{Fore.RESET}\n"
+        )
+        results = _check_for_results_folder(output)
+        results.mkdir()
+        [shutil.move(f, results / f.name) for f in output.iterdir() if f != results]
```
tkkuehn, line 230: Stylistically, I think I would prefer if this were a normal for loop because we're not doing anything with the list being generated by this list comprehension.

snakebids/output.py
```diff
@@ -0,0 +1,472 @@
...
+
+        # Move .snakebids file back to the top level
+        shutil.move(results / ".snakebids", output / ".snakebids")
+        _copy_snakemake_app(src, output, False)
+        return results
+    
+    # Convert to Bidsapp mode
+    if not force:
+        raise RunError(
```
tkkuehn, line 239: I think it would be a little nicer not to assume at this level that this exception will always result from a CLI invocation. I'd prefer this exception message to have a little less detail (specifically about the CLI invocation), then to catch it at a higher level and return the details of the required flags (this might require a custom-defined `ConversionError` or similar.

tkkuehn (changes requested):
Okay, finished taking a bit of a deeper look. From a high level I think this looks good and don't have anything to add to what @akhanf said, but I have a number of nitpicky comments. One other thing is that this could use a run of black (I forget whether we have that implemented as a GitHub action on this repo, so disregard if not necessary. Thanks for putting this PR together!

snakebids/output.py
pvandyken: For sure, I did it this way because I thought it would be easier to iterate over the directory items and filter out what I want, but looking again it actually adds very little ease.

snakebids/output.py
pvandyken: I might be conufused, but `RunError` is a custom error (see `snakemake/exceptions.py`). It does sound kind of generic though. I'm open to changing the name or describing a few more types.

There was a bit more I had meant to comment about Exception handling, which is sort of incomplete in this PR, but it's more general so I'll make a non-threaded comment.

pvandyken:
@tkkuehn raised an issue reminding me that my exception handling is rather incomplete in this PR. Most of the Exceptions raised, especially `RunError`, are predictable responses to user mistakes, so we don't nee a whole messy stack trace. Pretty simple fix: just need to catch `RunError` and print the message. But if there's any other opinions on how to handle these, I'm open to them.

pvandyken:
> 1. In bidsapp mode, I don't think it is right to have the`.snakemake` folder exist in the original workflow folder.
> 
> 
> It unnecessarily ties runs of the same app, but on different datasets and different output folders. E.g. say I install a snakebids app run.py as a console script, let's say it is hippunfold. Then if I wanted to run hippunfold on two completely independent datasets:
> 
> ```
> hippunfold bids_datasetA bids_datasetA/derivatives/hippunfold participant 
> hippunfold bids_datasetB bids_datasetB/derivatives/hippunfold participant 
> ```
> 
> These would share the same `.snakemake` folder (located wherever the workflow is installed), so would have issues with locking/unlocking the workflow, the `.snakemake` folder would unnecessarily know about files in both output folders, etc..
> 
> Also, depending on how he snakebids app was installed, the `.snakemake` folder could be either be tucked away in a virtualenv (not an obvious place for it), or if we have the entire app packaged in a single container (like hippunfold is), it would be in a read-only location (could perhaps be corrected with a bind path when running, but would be ad-hoc..).
> 
> So, I think it would be better to set the snakemake working directory to outputdir in bidsapp mode too.

I agree with that reasoning. The reason I didn't go that way in the first place was mainly because I was concerned with path resolution, which has been a problem for me in the past. Case in point: I have an atlas in my resources folder, the path to which is saved in my config. If we relocate the `cwd`, that relative path to my atlas becomes unresolved and we get an error. 

So if we want to move the cwd, we have to choose from a number of options:

1. We don't support relative paths in config files. Any path to a resource or whatever should be absolute to guarentee it will work. This isn't particularly friendly to our users and damages portability.
2. When in bidsapp mode, we go through the config, search for anything that may be a path and resolve it. This is messy and highly prone to error: how do we decide what a path is? What if the user doesn't want a path resolved, for whatever reason. etc.
3. We add more configuration/cmd-line args that allow the user to specify paths that should be resolved. This would increase Snakebids already rather bloated command-line/configuration interface and make configuration more confusing.
    - We did, in fact, take a similar approach with CLI args (allowing the user to specify them as Paths, then resolving them). But it's that "allow the user to specify them" that's important there, and the interface to do that was very natural (taking advantage of `configargparse`'s existing API). Such is not the case for any arbitrary item in the config.

Admittedly, the problem technically still exists in Workflow mode, but it doesn't seem as bad to me in that context because at least the entire snakebids app is being copied over, so resources will be safe. 

So, acknowledging the benefits of moving the cwd in bidsapp mode, as outline above, do they outweigh the negative effects of choosing one of the three options above (or, all too likely, is there a fourth option I'm not considering)?


> 1. If we make that change, then in bidsapp mode there is an issue with resolving the configfile path in the snakemake workflow, e.g. `configfile: config/snakebids.yml` won't be found since relative to the working directory we would have `code/snakebids.yml`.
> 
> 
> A fix for this could be to use snakemake's (somewhat recently added?) ability to get paths relative to the Snakefile:
> 
> ```
> configfile: workflow.source_path('../config/snakebids.yml')
> ```

The config file in the parent snakebids app is incomplete, we're not writing anything to it. So as is, just referencing it won't work, unless I'm missing something. We're already specifying a `configfile` in our call to `snakemake.main`, why not just pass the path to `./code/[config.yml]`?

>  1. Unrelated to the above, instead of having workflow authors use `bids=(root=config['output_dir'] ...`, which results in absolute file paths for all the workflow outputs, I think it would be better to have a new config variable that is either set to `''` (when in bidsapp mode) or `'results'` (when in workflow mode).. This retains the use of relative file paths for any workflow outputs, which is what snakemake recommends in general (for portability). I've gone ahead and made `config['root']`  for this, but open to suggestions on the naming..

I do prefer the `root` syntax, as it's more in line with our communications so far around this. And yes, this would make the child app more portable: currently if the user moves it it won't work. 


> I've made these changes in my commit, along with edits to test cases to accommodate the changes. I've also updated the project_template snakebids app to comply too. Note, we should perhaps add some test cases that spawn and use this template too.

Agreed. Also, some E2E tests of some sort would be a good idea.

pvandyken:
> Okay, finished taking a bit of a deeper look. From a high level I think this looks good and don't have anything to add to what @akhanf said, but I have a number of nitpicky comments. One other thing is that this could use a run of black (I forget whether we have that implemented as a GitHub action on this repo, so disregard if not necessary. Thanks for putting this PR together!

I was planning on running everything through black etc once we get this PR merged and work out #63, then do everything at once.

pvandyken:
Oh, one other consideration I had meant to mention:

All of the conversions are not very failure proof. If the process gets interrupted, files will be left in a half-copied state. I have some ideas to improve that, but they would be a bit of work to implement, so I just want feedback on how important this is. Can it wait for a future PR?

akhanf:
> I agree with that reasoning. The reason I didn't go that way in the first place was mainly because I was concerned with path resolution, which has been a problem for me in the past. Case in point: I have an atlas in my resources folder, the path to which is saved in my config. If we relocate the `cwd`, that relative path to my atlas becomes unresolved and we get an error.
> 
> So if we want to move the cwd, we have to choose from a number of options:
> 
>     1. We don't support relative paths in config files. Any path to a resource or whatever should be absolute to guarentee it will work. This isn't particularly friendly to our users and damages portability.
> 
>     2. When in bidsapp mode, we go through the config, search for anything that may be a path and resolve it. This is messy and highly prone to error: how do we decide what a path is? What if the user doesn't want a path resolved, for whatever reason. etc.
> 
>     3. We add more configuration/cmd-line args that allow the user to specify paths that should be resolved. This would increase Snakebids already rather bloated command-line/configuration interface and make configuration more confusing.
>        
>        * We did, in fact, take a similar approach with CLI args (allowing the user to specify them as Paths, then resolving them). But it's that "allow the user to specify them" that's important there, and the interface to do that was very natural (taking advantage of `configargparse`'s existing API). Such is not the case for any arbitrary item in the config.
> 
> 
> Admittedly, the problem technically still exists in Workflow mode, but it doesn't seem as bad to me in that context because at least the entire snakebids app is being copied over, so resources will be safe.
> 
> So, acknowledging the benefits of moving the cwd in bidsapp mode, as outline above, do they outweigh the negative effects of choosing one of the three options above (or, all too likely, is there a fourth option I'm not considering)?

The fourth option, and I believe the preferred way to resolve paths relative to the Snakefile is to use the `workflow.source_path()` function. The workflow author would apply this whenever it is accessing a file that is known to be relative to the workflow's Snakefile, and this would resolve appropriately in bidsapp mode or in workflow mode. 

E.g. in your case, you would have:

config.yml:
```
  my_atlas: '../resources/myatlas.nii.gz'  #relative to Snakefile
```

and in your workflow:
```
  my_atlas_resolved = workflow.source_path(config['my_atlas'])
```

This would resolve correctly regardless of what cwd was, unless you have any counter-examples?

Note: in some of my workflows (notably hippunfold) I used `config['snakemake_dir']`  to try to resolve things in resources, but the better solution is to use the Snakemake `workflow.source_path()` function to do this, I think I was either unaware of it or it didn't exist at the time..

> 
> > 1. If we make that change, then in bidsapp mode there is an issue with resolving the configfile path in the snakemake workflow, e.g. `configfile: config/snakebids.yml` won't be found since relative to the working directory we would have `code/snakebids.yml`.
> > 
> > A fix for this could be to use snakemake's (somewhat recently added?) ability to get paths relative to the Snakefile:
> > ```
> > configfile: workflow.source_path('../config/snakebids.yml')
> > ```
> 
> The config file in the parent snakebids app is incomplete, we're not writing anything to it. So as is, just referencing it won't work, unless I'm missing something. We're already specifying a `configfile` in our call to `snakemake.main`, why not just pass the path to `./code/[config.yml]`?

The `configfile: ...` in the Snakefile is needed so it can find the configfile when running in workflow mode with the snakemake CLI. Unless you are doing something different from what is in the project template Snakefile boilerplate?

Yes when running in bids app mode, the parent (incomplete) config file gets ultimately overridden by run.py with `--configfile`. But if we used `configfile: ./code/config.yml` in the Snakefile then the workflow mode would have the wrong path.  

So to be clear,
bidsapp mode:  
- run.py -> sources `configfile: ` in Snakefile from parent (uncustomized) -> sources overriding config specified with `--configfile`
workflow mode:
- snakemake -> sources `configfile: ` in Snakefile in current folder

Note: the behaviour for bidsapp mode above is how snakebids apps (at least mine) have been using up till now. Not to say that we should continue for that reason, but to me I don't see how we can have a bidsapp mode without setting the cwd to the output folder.

pvandyken:
> The fourth option, and I believe the preferred way to resolve paths relative to the Snakefile is to use the `workflow.source_path()` function. The workflow author would apply this whenever it is accessing a file that is known to be relative to the workflow's Snakefile, and this would resolve appropriately in bidsapp mode or in workflow mode.
> 
> E.g. in your case, you would have:
> 
> config.yml:
> 
> ```
>   my_atlas: '../resources/myatlas.nii.gz'  #relative to Snakefile
> ```
> 
> and in your workflow:
> 
> ```
>   my_atlas_resolved = workflow.source_path(config['my_atlas'])
> ```
> 
> This would resolve correctly regardless of what cwd was, unless you have any counter-examples?

Excellent. I like this.


> The `configfile: ...` in the Snakefile is needed so it can find the configfile when running in workflow mode with the snakemake CLI. Unless you are doing something different from what is in the project template Snakefile boilerplate?
> 
> Yes when running in bids app mode, the parent (incomplete) config file gets ultimately overridden by run.py with `--configfile`. But if we used `configfile: ./code/config.yml` in the Snakefile then the workflow mode would have the wrong path.
> 
> So to be clear, bidsapp mode:
> 
>     * run.py -> sources `configfile: ` in Snakefile from parent (uncustomized) -> sources overriding config specified with `--configfile`
>       workflow mode:
> 
>     * snakemake -> sources `configfile: ` in Snakefile in current folder
> 
> 
> Note: the behaviour for bidsapp mode above is how snakebids apps (at least mine) have been using up till now. Not to say that we should continue for that reason, but to me I don't see how we can have a bidsapp mode without setting the cwd to the output folder.

I'm still confused here. If you set configfile in the `Snakefile` as such:
```
configfile: config/[config.yaml]
```
it will still work properly in bidsapp mode because it will be overriden via the `--configfile` argument. In fact, when running using `run.py`, the `configfile: ...` declaration in the `Snakefile` is never taken into account, so that's only important when directly using the Snakemake CLI. And that situation, as I understand, is only relevant in a "workflow mode" context. So can't you just leave `configfile: ...` as above, to account for Snakemake CLI, then let the Snakebids CLI determine the correct file in every other case?

akhanf:
> I'm still confused here. If you set configfile in the `Snakefile` as such:
> 
> ```
> configfile: config/[config.yaml]
> ```
> 
> it will still work properly in bidsapp mode because it will be overriden via the `--configfile` argument. In fact, when running using `run.py`, the `configfile: ...` declaration in the `Snakefile` is never taken into account, so that's only important when directly using the Snakemake CLI. And that situation, as I understand, is only relevant in a "workflow mode" context. So can't you just leave `configfile: ...` as above, to account for Snakemake CLI, then let the Snakebids CLI determine the correct file in every other case?

No, when I was testing things, if I had:

bids-app mode write the config file to `code/config.yml`, and the snakefile having `configfile: config/config.yml`, then snakemake would error out because it could not find `config/config.yml` relative to the working directory.

Alternatively (how it was set-up prior to this PR), if bids-app mode writes the config file to `config/config.yml`, then it is ok. So it is the change to using `code/` to copy the bidsapp mode config files that necessitates the use of `workflow.source_path()` in the Snakefile.  You should be be able to try this out yourself to confirm.

akhanf:
Forgot to mention, this is because even if the `--configfile` CLI option is used, the `configfile:` Snakefile statement is still evaluated, bu then config is overwritten with whatever is provided in CLI.

pvandyken:
> Forgot to mention, this is because even if the `--configfile` CLI option is used, the `configfile:` Snakefile statement is still evaluated, but then config is overwritten with whatever is provided in CLI.

Ah, now I understand. Okay, then your solution is definitely reasonable.

pvandyken:
It should be noted that, due to the changes made by @akhanf, this PR will be BREAKING for existing workflows. Anyone adopting the new version will need to update their workflow files with the `workflow.source_path()` syntax, or bidsapp mode won't work.

akhanf:
We should probably instead start merging this into a 0.4.0 branch rather than master as you suggested I think in another PR..

pvandyken:
I'm not sure if a new branch would give too much advantage (unless people are in the habit of installing directly off the master branch), but I will manually update the version of this pull to `0.4.0-dev.1`, and disable bumpversion for now until we can get our versioning system under control.

In the meantime, I'm trying to summarize the issues yet to be resolved at the top of this PR. I think we've worked out handling of old `.snakemake` folders: just keep whatever `.snakemake` folder we find in the output folder (i.e. do nothing).

Unless I'm told otherwise (@tkkuehn), I'll just adjust error handling as I suggested above (catch `RunError` and print the message). 

Other than the listed todos, is there anything else to be considered for this PR? (@akhanf @tkkuehn @kaitj )

kaitj:
I haven't had a chance to go through this PR, but everything discussed seems reasonable to me. 

I second a 0.4.x branch at this point if it will automatically publish a release to `pypi`, given this seems to be a much larger breaking change. @pvandyken and I had a brief discussion about automated versioning / release publishing and how to best handle that should be discussed at the dev lunches. Once there is some consensus on that and we have an implementation, it may be fine to merge subsequent changes directly to master moving forward.

pvandyken:
Okay, so I've finished all the checkboxes above. The sphinx docs are written, but I thought I'd leave them for the next PR as this one is getting a bit long. 

I've REBASED the PR, so if you were working from the old one, please transfer to the latest version (or just wait till it's all pushed into 0.4.0). If this causes problems, I still have a backup of the old tree saved. 

So just to summarize what I think were the major issues:
1. `cwd` will always be in the output folder. Snakebids will **never** touch the `.snakemake/` folder, either for copying or converting an app/output. 
2. `RunErrors` are caught, and their message is printed without trace.
3. No special proofing for interruptions, copy failures, etc. I consider this a problem, but would like to address it in a future PR.

In total, I believe this is good enough for an alpha version: we can work out the remaining kinks in future PRs. So if you all are content (@akhanf @tkkuehn @kaitj ), I'll go ahead and merge into `0.4.0`

tkkuehn:
Looks good to me!

PR #64, opened 2021-11-10
pin docutils<0.18 for docs

akhanf:
should fix failed readthedocs builds


akhanf:
https://github.com/readthedocs/readthedocs.org/issues/8616

pvandyken:
Ah, I was going to ask about that, glad to see a fix. I'll make sure the pyproject gets similarly updated when we fully switch over

PR #65, opened 2021-11-10
Project template updates

akhanf:
- workflow files in main dir now (only run.py in subfolder)
- use pathlib instead of os.path
- replaced _py cookiecutter variable with just py, to ease readability

To do:
 - [ ] update any documentation that refers to this <-- couldn't find anything that needed updating explicitly..

This isn't really a breaking change for snakebids, but putting it in 0.4.0 in any case..


akhanf:
I also moved the project template folder to the root folder as discussed earlier today..

pvandyken:
Looks good for merging!

pvandyken (approved):

Issue #63, opened 2021-11-05
Code Quality checks


pvandyken:
On implementing poetry, it would be good to agree on what formatters/quality checkers should be used in development. This would let us set up checks in gh-actions to ensure code quality. 

Currently we have:
- black
- pylint

Other possibilities:
- [isort](https://pypi.org/project/isort/) - Automatically organizes imports first by (stdlib, 3rd party, local) then by alphabet. I like it to keep things organized.
- [flake8](https://flake8.pycqa.org/en/latest/) - Stricter than black, but I don't think it does automatic formatting (so it's often combined with black). It depends how picky we want to be.

I'm also using [pre-commit](https://pre-commit.com/) in one of my other projects and find it quite useful. It basically just runs things like black, lint checks, etc on every commit. It's nice from a dev perspective because it can be set up with just one command, so you don't need to remember to do a bunch of checks every time you push. Happy to set it up for snakebids if we're interested.

kaitj:
Most of my recent experience has been with `black`, though I have used `pylint` and `flake8` in the past. If I recall from past discussions, `black` doesn't fully follow PEP8, but is more user-readable. Any is fine with me as long as formatting is consistent across different contributions.

`isort` seems like a good thing to have. I also like to have the library imports organized, though I haven't used anything like this yet.

I would say the `pre-commit` is definitely nice to have, with a secondary check via `gh-actions`. In `gh-actions`, for security purposes, they don't allow commits to be pushed to forks (unless something has changed with the git tokens), which eliminates the automatic formatting in those cases. We have something similar setup for the afids-validator, with a pre-commit and a gh-actions.

pvandyken:
Ya, I know when combining flake8 and black you need to make a few flake8 exceptions for them to be in agreement.

akhanf:
Those suggestions sound good to me!

tkkuehn:
I use pylint generally, and it is pretty picky but generally works well with black (I can't think of any conflicts off the top of my head).

PR #68, opened 2021-11-11
Integrate pre-commit hooks and gh-actions linting

pvandyken:
This PR contains the new CI flow for snakebids. It currently incorporates `black`, `pylint`, `isort`, and `flake8`, as discussed in #63. I went through all the code and refactored/reformatted to give us a clean slate for starting.

I customized a few of the linters: `flake8`, notably, needs a few tweaks to work with `black`. I also added a few exceptions to `pylint`, and set it to fully ignore `test_*` files, as it was raising way too many issues about regular practices (e.g. using pytest fixtures, using private methods). I've never used `pylint` in my own projects, so I'd love to hear if you have a better way
of addressing this problem.

I installed a task runner: [poethepoet](https://github.com/nat-n/poethepoet). I've found it to be a nice way to standardize common tasks. Basically it lets you define custom tasks in your `pyproject.yaml` under `[tool.poe.tasks]`. These tasks can then be run via:
```
poetry run poe task-name
```
Or, you can install `poethepoet` globally via `pipx`: `pipx install poethepoet`. Poe will then autodetect your poetry env so that you can just call:
```
poe task-name
```
when in the project directory.

All of the linters will run as a pre-commit hook. You just have to run `pre-commit install` or `poe setup` before your first commit. In addition to the linters, it packs a few extras that check for yaml mistakes, trailing whitespace, and a few other things.

All the linters are run as a `gh-action`, so we'll get code quality reports on PRs.

Lastly, this PR address the comments made in #60 regarding the `pyproject.toml`. All the `>=` version specifiers have been changed to `^`, except for Python, where we do support (at least, we explicitely test) every version from 3.7 up to 3.10, and Snakemake. The snakemake dependency is currently at `>=5.28.0`, which is a pretty low version. @akhanf, does the new `workflow.source_path()` syntax come in a later version? I know you mentioned it was recent. Other than that, I don't think it particularly matters what Snakemake version is used.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,10 @@
+[flake8]
+extend-ignore = E203,W503
+max-line-length = 88
+extend-exclude =
+    snakebids/tests/
+    .pytest_cache/
+    dist/
+    .venv/
+    .hypothesis/
+    docs/
--- a/.github/workflows/test_deploy.yml
+++ b/.github/workflows/test_deploy.yml
@@ -8,8 +8,65 @@ on:
   pull_request:
 
 jobs:
+  quality:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Clone repo
+        uses: actions/checkout@v2
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.9'
+      - uses: actions/cache@v2
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-3.9
+          restore-keys: ${{ runner.os }}-pip-3.9
+      #----------------------------------------------
+      #  -----  install & configure poetry  -----
+      #----------------------------------------------
+      - name: Install Poetry
+        uses: snok/install-poetry@v1
+        with:
+          virtualenvs-create: true
+          virtualenvs-in-project: true
+      #----------------------------------------------
+      #       load cached venv if cache exists
+      #----------------------------------------------
+      - name: Load cached venv
+        id: cached-poetry-dependencies
+        uses: actions/cache@v2
+        with:
+          path: .venv
+          key: venv-${{ runner.os }}-${{ hashFiles('**/poetry.lock') }}-3.9
+      #----------------------------------------------
+      # install dependencies if cache does not exist
+      #----------------------------------------------
+      - name: Install dependencies
+        if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
+        run: poetry install --no-interaction --no-root
+      #----------------------------------------------
+      # install your root project, if required
+      #----------------------------------------------
+      - name: Install library
+        run: poetry install --no-interaction
+
+      #----------------------------------------------
+      #       run python style checks
+      #----------------------------------------------
+      - name: isort
+        run: poetry run isort snakebids -c
+      - name: Black
+        run: poetry run black snakebids --check
+      - name: Flake8
+        run: poetry run flake8 snakebids
+      - name: pylint
+        run: poetry run pylint snakebids
+
   test:
     runs-on: ubuntu-latest
+    needs: [ 'quality'  ]
     strategy:
       matrix:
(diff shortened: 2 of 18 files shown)


kaitj:
Overall, looks good to me. Haven't had a chance to run the actions yet to see if any errors come up - will do that later this morning before completing the review.

> All of the linters will run as a pre-commit hook. You just have to run pre-commit install or poe setup before your first commit. In addition to the linters, it packs a few extras that check for yaml mistakes, trailing whitespace, and a few other things.

Thinking of more things that would be good to have within the repo :sweat_smile:. We might want to include a contributor's doc with all these instructions along with setting up the dev environment for future contributors or anyone not as familiar with everything that is being set up.

> All the linters are run as a gh-action, so we'll get code quality reports on PRs.

I can't recall at the moment, but was the plan to also have this action make any corrections or just inform the user? If the former, I don't think any changes are currently pushed back to the branch, otherwise can ignore this Q.

> The snakemake dependency is currently at >=5.28.0, which is a pretty low version.

Given this is an intiial PR to get all these things setup, I'd be okay with to just have the versions as is and have a future PR update the snakemake version once someone has given this a test to make sure nothing is broken.

kaitj (commented):
Just another comment, but otherwise to me, it looks good. 

Noticed the dependencies in the actions was not cached between jobs despite the setup for caching. Not sure if this was intended (see quality vs test jobs - both use the same keys). This is also true across the different python versions. Likely to do with the fact the multiple jobs can run concurrently. 

I think one possible way to get around this is to make the quality and test jobs dependent on another (similar to `test-deploy`) and also include python version to the key (to make each cache version specific).

akhanf:
Looks great at a glance! Had a thesis external report deadline sneak up on
me so will have to put my head down today to complete it, but will try to
respond later this evening.

pvandyken:
@kaitj 
> Thinking of more things that would be good to have within the repo 😅. We might want to include a contributor's doc with all these instructions along with setting up the dev environment for future contributors or anyone not as familiar with everything that is being set up.

Ya for sure. I don't think there's a lot of things to note, just the steps to install and test everything. Anything else I'm missing? Is it better to have a separate doc, or just a section in the `README`? 

> I can't recall at the moment, but was the plan to also have this action make any corrections or just inform the user? If the former, I don't think any changes are currently pushed back to the branch, otherwise can ignore this Q.

Based on other examples I've seen of similar workflows, it looks like it's more typical to just do a check (e.g., that's how Snakemake does it). So I have both `black` and `isort` set to not change anything but error out if they would change something.

pvandyken:
@tkkuehn 
> Noticed the dependencies in the actions was not cached between jobs despite the setup for caching. Not sure if this was intended (see quality vs test jobs - both use the same keys). This is also true across the different python versions. Likely to do with the fact the multiple jobs can run concurrently.
> 
> I think one possible way to get around this is to make the quality and test jobs dependent on another (similar to `test-deploy`) and also include python version to the key (to make each cache version specific).

So I believe the cache works across job contexts (i.e. when making a new push). But yes, the same environment is setup twice for quality and for checks currently. So I'll add the dependency and specify the python version in the key as suggested.

kaitj:
> Ya for sure. I don't think there's a lot of things to note, just the steps to install and test everything. Anything else I'm missing? Is it better to have a separate doc, or just a section in the `README`?

A section is probably enough, we can look into adding `CONTRIBUTING` later on if necessary

> Based on other examples I've seen of similar workflows, it looks like it's more typical to just do a check (e.g., that's how Snakemake does it). So I have both black and isort set to not change anything but error out if they would change something.

Sounds good. That is probably the best way to handle it anyways so as to not have to deal with tokens in PRs from forks.

pvandyken:
Just added a contributer section. If that looks good (@kaitj), I think this one's good for merging.

akhanf (approved):

kaitj (approved):
Looks good to me!

PR #75, opened 2021-11-17
Setup release drafter

### LINKED ISSUE #72: setup release-drafter
Set up release-draft actions to collect and automatically update changelog

kaitj:
This PR (based on `0.4.0`) starts to incorporate the `release-drafter` into the github actions workflow. This adds the following functionality. 

- [x] Resolve version by label
- [x] Automatic version bumping using resolved version (uses existing `bump_version.yml` workflow)
- [x] Add missing labels

The labels (if missing or there are others we wish to include) still need to be added to the repo.

Resolves #72. Note, the `release-drafter` won't actually start to work until it is merged into `main` and will cause the CI for this branch to fail until merged.

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,0 +1,43 @@
+name-template: '$RESOLVED_VERSION'
+tag-template: 'v$RESOLVED_VERSION'
+template: |
+  ## Changes
+
+  $CHANGES
+categories:
+  - title: '🚀 Features'
+    labels:
+      - 'breaking'
+      - 'enhancement'
+  - title: '🐛 Bug Fixes'
+    labels:
+      - 'bug'
+  - title: '🧰 Maintenance'
+    labels:
+      - 'maintenance'
+      - 'test'
+  - title: '📝 Documentation'
+    labels:
+      - 'documentation'
+
+exclude-labels:
+  - skip_changelog
+
+change-template: '- $TITLE @$AUTHOR (#$NUMBER)'
+change-title-escapes: '\<*_&' # You can add # and @ to disable mentions, and add ` to disable code blocks.
+
+version-resolver:
+  # major:
+  #   labels:
+  #     - 'breaking'
+  minor:
+    labels:
+      - 'breaking'
+      - 'enhancement'
+  patch:
+    labels:
+      - 'maintenance'
+      - 'bug'
+      - 'test'
+      - 'documentation'
+  default: patch
--- a/.github/workflows/bump_version.yml
+++ b/.github/workflows/bump_version.yml
@@ -15,39 +15,42 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Find current version
-        id: version-check
-        env:
-          next_version: ${{ steps.release-drafter.outputs.name }}
+      - name: Get previous release version
         run: |
-          current_version=`grep 'version = *' pyproject.toml | sed 's/version = //' | tr -d '"'`
-          echo "\nCurrent version: $current_version"
-          echo "Next resolved version: $next_version"
-          if [[ $next_version == $current_version ]]; then
-            echo '::set-output name=commit::false'
+          echo "PREV_VER=$(cat pyproject.toml | grep -o -E '(version\s=\s)([[:punct:]])([0-9]+.[0-9]+.[0-9]+.+)([[:punct:]])' | cut -d ' ' -f 3 | tr -d '"')" >> $GITHUB_ENV
+
+      - name: Get previous bump version
+        env:
+          PREV_VER: ${{ env.PREV_VER }}
+        run: |
+          if [[ "$PREV_VER" != *"-prerelease."* ]]; then
+            echo "OLD_BUMP=0" >> $GITHUB_ENV
           else
-            echo '::set-output name=commit::true'
+            echo "OLD_BUMP=$(echo $PREV_VER | cut -d '.' -f 4)" >> $GITHUB_ENV
           fi
+
+      - name: Bump version
+        env:
+          BUMP_VER: ${{ env.OLD_BUMP }}
+        run: |
+          echo "NEW_BUMP=$(($BUMP_VER + 1))" >> $GITHUB_ENV
 
       - name: Update version in files
         uses: jacobtomlinson/gha-find-replace@master
-        if: steps.version-check.outputs.commit == 'true'
         with:
           include: 'pyproject.toml'
-          find: 'version = "[0-9]+.[0-9]+.[0-9]+"'
-          replace: 'version = "${{ steps.release-drafter.outputs.name }}"'
+          find: 'version = "(?:([0-9]+.[0-9]+.[0-9]+.+)|([0-9]+.[0-9]+.[0-9]+))"'
+          replace: 'version = "${{ steps.release-drafter.outputs.name }}-prerelease.${{ env.NEW_BUMP }}"'
 
       - name: Commit updates
-        if: steps.version-check.outputs.commit == 'true'
         env:
-          SNAKEBIDS_VERSION: ${{ steps.release-drafter.outputs.name }}
+          SNAKEBIDS_VERSION: ${{ steps.release-drafter.outputs.name }}-prerelease.${{ env.NEW_BUMP }}
         run: |
           git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
           git config --local user.name "github-actions[bot]"
-          git commit -m "Bump version to $SNAKEBIDS_VERSION" -a
+          git diff-index --quiet HEAD || git commit -m "Bump version to $SNAKEBIDS_VERSION" -a
 
       - name: Push changes
-        if: steps.version-check.outputs.commit == 'true'
         uses: ad-m/github-push-action@master
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/bump_version.yml
+++ b/.github/workflows/bump_version.yml
@@ -17,7 +17,7 @@ jobs:
 
       - name: Get previous release version
         run: |
-          echo "PREV_VER=$(cat pyproject.toml | grep -o -E '(version\s=\s)([[:punct:]])([0-9]+.[0-9]+.[0-9]+.+)([[:punct:]])' | cut -d ' ' -f 3 | tr -d '"')" >> $GITHUB_ENV
+          echo "PREV_VER=$(cat pyproject.toml | grep -o -E '(version\s=\s)([[:punct:]])([0-9]+\.[0-9]+\.[0-9]+.+)([[:punct:]])' | cut -d ' ' -f 3 | tr -d '"')" >> $GITHUB_ENV
 
       - name: Get previous bump version
         env:
@@ -39,7 +39,7 @@ jobs:
         uses: jacobtomlinson/gha-find-replace@master
         with:
           include: 'pyproject.toml'
-          find: 'version = "(?:([0-9]+.[0-9]+.[0-9]+.+)|([0-9]+.[0-9]+.[0-9]+))"'
+          find: 'version = "(?:([0-9]+\.[0-9]+\.[0-9]+.+)|([0-9]+\.[0-9]+\.[0-9]+))"'
           replace: 'version = "${{ steps.release-drafter.outputs.name }}-prerelease.${{ env.NEW_BUMP }}"'
 
       - name: Commit updates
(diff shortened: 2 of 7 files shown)


kaitj:
@akhanf @tkkuehn @pvandyken This isn't quite ready for a review yet, but wanted some discussion on setting this up. Can go over this PR in more detail on Friday as well. We talked about some of these things during the last dev meeting. I currently have it using the the existing workflows with `release-drafter` to do version bumping.

I think the old version bump use to increment the version for each PR (not sure if it had to be released)? `release-drafter` relies on the previous release tag to determine the next release version. This release has to be manually triggered (the way it is currently set up). As a result, each accumulated PR will not necessarily bump the version (if that makes sense), but prioritizes the dominant version (`major` > `minor` > `patch`) and will only bump once (eg. 2 `patch` PRs in the same release would only be a single version bump; 0.4.0 -> 0.4.1). One question was whether we are still going to be minting a new release every PR (if yes, this isn't as much of a concern).

Also wasn't familiar with the `deploy` part of the `test_deploy` workflow, and whether this part of the action would get triggered upon release or if we need to have a separate release trigger.

.github/workflows/bump_version.yml
```diff
@@ -8,13 +8,30 @@ jobs:
...
-        uses: jaumann/github-bumpversion-action@v0.0.7
+
+      - name: Update changelog
+        uses: release-drafter/release-drafter@v5
+        id: release-drafter
         env:
-          GITHUB_TOKEN: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Update version in files
```
pvandyken, line 18: It looks like this will run and push a new commit on every PR? Is there anyway to first check and see if the version needs updating, so that we don't end up with a bunch of bump commits that don't do anything?

.github/release-drafter.yml
```diff
@@ -0,0 +1,36 @@
...
+change-template: '- $TITLE @$AUTHOR (#$NUMBER)'
+change-title-escapes: '\<*_&' # You can add # and @ to disable mentions, and add ` to disable code blocks.
+
+version-resolver:
+  major:
+    labels:
+      # - 'breaking'
+  minor:
+    labels:
+      - 'breaking'
```
pvandyken, line 30: Do we not want `breaking` to be a major? Or are we waiting for `1.0` to start before we do that (which makes total sense)?

kaitj: I think we probably want to wait until `1.0` before moving this label back to major (to avoid any accidental major bump in the meantime). I tried to look at [semver](https://semver.org/) for some guidance on this, but I didn't see anything on pre-public release (`0.x.x`), so this was what I ended up going with.

.github/workflows/bump_version.yml
kaitj: I've left the workflow triggers the same, though that said, the version won't necessarily bump every PR merge (unless there is a release). I can show an example tomorrow as I've been testing this on a fork (also potentially live demo if it will help clear things up).

kaitj: Actually re-reading this, I see what you mean. We can definitely set a conditional flag to check if the versions are different, and only run the commit if they are.

.github/release-drafter.yml
pvandyken: Yeah, I completely agree, just thought it would be good to be explicit

kaitj: Major bumps are disabled (and tested) as of fe264f7.

.github/workflows/bump_version.yml
kaitj: 5983384 introduces checking for current vs resolved version - a commit will only occur if the versions differ.

pvandyken: Ah, very true. Thanks for looking into this nonetheless

.github/release-drafter.yml
```diff
@@ -0,0 +1,36 @@
...
+  $CHANGES
+categories:
+  - title: '🚀 Features'
+    labels:
+      - 'breaking'
+      - 'enhancement'
+  - title: '🐛 Bug Fixes'
+    labels:
+      - 'bug'
+  - title: '🧰 Maintenance'
```
pvandyken, line 15: Is the idea for maintenance that it's meant for automated tasks (like bump version)? If so, is it good to include it in the changelog?

pvandyken: Although, just further thinking this through, if maintenance *is* for automated tasks, then we should have another label for refactoring, add tests, or other pushes that don't change or fix anything but improve code stability. Or just use maintenance for all of that

pvandyken: And now I'm realizing that automated tasks is really only relevant for commit tags, and here we're discussing PR tags. So just ignore my above points :)

kaitj: Haha, well the template is very flexible to however we want to set it up, so for now I've tried to leave it as general as possible until there is some discussion on those labels. The original idea of maintenance, in context to afids, was for any dependency version bumps and as you mentioned, refactoring, etc. I think by default, the release-drafter uses a label defined as "chore" and not "maintenance".

We should figure out what kind of labels we would like and how they map to different version bumps.

kaitj:
This should be ready for review now - I've added in the missing labels that are currently listed in the `release-drafter.yml`. Any new labels can be added in later.

.github/workflows/test_deploy.yml
```diff
@@ -113,6 +113,17 @@ jobs:
       run: |
         poetry run pytest --doctest-modules --ignore=docs --ignore=project_template
 
+  draft_changelog:
```
pvandyken, line 116: Including this here would cause the changelog to be updated on every *pull request*. But this will cause problems if we want to make a release with pending PRs (which will happen more often than not, I expect). So I think having this just in the `bump_version.yml` is sufficient?

.github/workflows/test_deploy.yml
```diff
@@ -113,6 +113,17 @@ jobs:
...
+    needs: [ 'test' ]
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Draft and update change log
+        uses: release-drafter/release-drafter@v5
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
   deploy:
```
pvandyken, line 127: I brought this up in #80, but deploy should be moved to its own workflow that runs whenever we publish a release. Right now, this will run on every PR to main, especially PRs that are already tagged with a version.

.github/release-drafter.yml
```diff
@@ -0,0 +1,36 @@
...
+  # major:
+  #   labels:
+  #     - 'breaking'
+  minor:
+    labels:
+      - 'breaking'
+      - 'enhancement'
+  patch:
+    labels:
+      - 'maintenance'
```
pvandyken, line 34: It might be useful to add `documentation` and `tests` as labels right of the bat? I think they would be useful for organization purposes.

pvandyken:
Overall looks really good. Thanks so much for your work here @kaitj

.github/workflows/test_deploy.yml
kaitj: That works for me - i just tried to get a setup of the drafter here. There is a release actions trigger that we can use. Happy to move this as part of this PR or in a separate PR

.github/workflows/test_deploy.yml
kaitj: Sounds good to me. Would that mean any changes we would want to make to the CHANGELOG we do manually since it will now only update on a merge to master

.github/release-drafter.yml
kaitj: Added these two labels in 5c5abb9

.github/workflows/test_deploy.yml
kaitj: Took this out of the `test_deploy` workflow in 75969ed.

pvandyken: By `CHANGELOG` do you mean the release changelog? Or the file `CHANGELOG`. If the file, I didn't realize that was even touched by this. For the release changelog, I believe it's sufficient for it to be updated upon merge, no? But I think I'm missing something here.

.github/workflows/test_deploy.yml
pvandyken: The python matrix is not necessary. Python is being installed solely for pushing to `PyPI`, so any version will do

.github/workflows/test_deploy.yml
kaitj: Ahh sorry, I meant release changelog. I was just thinking of the case where it gets merged in and then we want to make a change to the PR title - shouldn't actually be an issue though.

.github/workflows/test_deploy.yml
kaitj: Sounds good. 

The deploy currently in this checks for the tag - in the updated local version, I've switched to just checking out that latest tag. Haven't pushed these changes yet though.

tkkuehn (approved):
I think this looks ready to merge at this point. Thanks for figuring all this out!

akhanf (approved):
Super! Thanks for putting this together!

kaitj:
Marked as draft as we want to add pre-release versioning as discussed during the dev lunch

kaitj:
Added in the prerelease version bumping - had to do it in a few steps with how actions workflows handles variables and expressions. Workflow will look for `prerelease.x` in the version, and if it doesn't find it, sets it to `prerelease.1`, otherwise increments the prerelease version each time a PR / push to main occurs (eg `x.4.x.prerelease-1` -> `x.4.x.prerelease-2`). It **does not** create a new tag.

Also took advantage of the PR to split the test and deploy workflow (which can be triggered manually via actions). Showed it briefly during the last dev lunch, but under the actions in the deploy workflow, there will be `Run workflow` drop down option once this PR is merged (similar as in the screenshot). There are some required inputs to ensure the workflow isn't just accidentally run and has the same permissions as publishing it from the draft (user needs to have the correct repo permissions). We can add any input fields we'd like to this drop down.

image

akhanf (approved):

.github/workflows/bump_version.yml
```diff
@@ -8,13 +8,50 @@ jobs:
...
+
+      - name: Update changelog
+        uses: release-drafter/release-drafter@v5
+        id: release-drafter
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Get previous release version
+        run: |
+          echo "PREV_VER=$(cat pyproject.toml | grep -o -E '(version\s=\s)([[:punct:]])([0-9]+.[0-9]+.[0-9]+.+)([[:punct:]])' | cut -d ' ' -f 3 | tr -d '"')" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
kaitj: Ahh you might be right. I think in this case, I think it would match any character - will run a quick test later today and update if needed. Haven't really tried the regex here on anything but a version number.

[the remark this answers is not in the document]
kaitj: Regex now updated to match literal `.` in both bump and deploy workflows via 1cd5d61 and 637473c

akhanf:
@kaitj feel free to merge this in whenever you're ready!

akhanf:
actually -- one more thing to consider now -- 
Now that I added branch protection, I remember why I removed it in the first place -- it seems the branch protection rules also apply to actions, e.g. so now the bumpversion action is failing when trying to push:
https://github.com/akhanf/snakebids/actions/runs/1523559717

Any suggestions? I guess we could just remove some of the branch protection on main..

kaitj:
> @kaitj feel free to merge this in whenever you're ready!

Sounds good, just waiting for @tkkuehn to have a chance to have another look at it before merging.

> actually -- one more thing to consider now --
> Now that I added branch protection, I remember why I removed it in the first place -- it seems the branch protection rules also apply to actions, e.g. so now the bumpversion action is failing when trying to push:
> https://github.com/akhanf/snakebids/actions/runs/1523559717
> 
> Any suggestions? I guess we could just remove some of the branch protection on main.. 

One idea is to use the "Restrict who can dismiss pull request reviews" option (not sure if this has to be in an org or if it can be a user repo) that looks like was added to Github a last week. If it is possible to set the user to the `github-actions[bot]`, can potentially rework the workflow to run a `git push` command instead of relying on the `ad-m/github-push-action` action.

There are also actions in the github marketplace that temporarily disable the branch protection during the workflow before re-enabling the protection after the workflow is completed. Would need to look at this a little more, but it's also an option.

tkkuehn (approved):
Looks good to me!

akhanf:
ok I've disabled branch protection altogether for now (I'll try harder to not make any goofs!)
since we have approvals I will merge this in so we can start drafting a new release

PR #84, opened 2021-11-26
upgrade to pybids-0.14.0

akhanf:
updated pyproject deps
removed extension_initial_dot (now set by default)
removed fix_dd console script

Note: while doing this (danger of multi-tasking) I ended up pushing this commit to main directly by accident.. I did a hard reset right away, although the version bump still happened.. Thus I have now added branch protection rules so this can't happen again.


kaitj (approved):

tkkuehn (approved):
Looks good to me!

Issue #86, opened 2021-12-06
Poetry doesn't like `x.x.x-prerelease.x` syntax


pvandyken:
I get an error when running `poetry install`. I've tested with `x.x.x-a.x` and `x.x.x-pre.x`, both of which work fine. @kaitj

kaitj:
That's good to know, I'll update the workflow tonight and make a PR

Issue #62, opened 2021-11-05
Noncanonical use of __init__.py


pvandyken:
Snakebids currently uses `__init__.py` as the file holding the main body of code. Although other apps do this (notably, ahem, snakemake), this is a noncanonical way of using `__init__.py`, which should typically only be used as a way of organizing imports into the module namespace so that it's easier to import things. (e.g. so that users can simply do `from snakebids import x` rather than `from snakebids.file import x`).

As such, I'd like to move most of the code in that file to `main.py`. Is this alright with the other maintainers?

akhanf:
Yes sounds good to me!

akhanf:
Yep that works for me! I probably learned from that bad example!

PR #87, opened 2021-12-06
Reorganize snakebids directory

pvandyken:
As discussed in #62, the `__init__.py` file should not be used to store code, but rather just for managing imports. This PR moves the code within the file to `bids.py`.

I added a new tool I like to use to manage `__init__.py` files: [`mkinit`](https://pypi.org/project/mkinit/). It allows you easily specify which modules should be imported into the `__init__.py` file using `__submodules__=[...]`. It takes care of all the code to accomplish that, putting it within the `<AUTOGEN_INIT>` tags. The tool can be called using the `poe mkinit` command, which I programmed into the `pyproject.toml` file. 

I also reorganized the main directory a bit, moving `outputs.py` and `snakemake_io.py` into a new `utils/` directory.

Finally, I ran `poetry update`, so anything that could be updated within our constraints was.

Note that I had to manually change the issue specified in `pyproject.toml` to prevent an error, as per #86. 

Address issue #62


akhanf (approved):
Looks good to me

kaitj (commented):
Overall looks good to me - just one question.

Is `pybids` still being imported anywhere other than in `snakebids/bids.py`? I only saw it imported in the testcases, which isn't a big deal. Just wouldn't want any confusion between `snakebids/bids.py` (relative import) and `pybids`.

There will be a merge conflict due to the version in `pyproject.toml`. I believe #88 bumped it to `0.4.0-pre.2` after the merge.

kaitj:
> @kaitj I'm open to changing the name of the file, although we already have a function within the now `bids.py` file called `bids()`. All of the functions in `bids.py` are imported into `__init__.py`, so they can still be directly imported into projects. Let me know if you have any suggestions

Don't know if we necessarily need to change it if `pybids` isn't being called elsewhere. Will think about this some more.

kaitj (approved):

pvandyken:
> Looks good to me. One question: Why aren't we including `app.py` in the init file?

@tkkuehn Simply because it wasn't included before. Happy to do so now (with the realization that it will make everything in app available under `snakebids.x`

tkkuehn:
> > Looks good to me. One question: Why aren't we including `app.py` in the init file?
> 
> @tkkuehn Simply because it wasn't included before. Happy to do so now (with the realization that it will make everything in app available under `snakebids.x`

Makes sense! I wasn't thinking about it that way but that seems totally reasonable. Feel free to merge from my end!

PR #88, opened 2021-12-06
Update prerelease version to use `pre`

kaitj:
This patch changes the prerelease version to use `x.x.x-pre.x` instead of `x.x.x-prerelease.x`. The latter doesn't seem to play nice with `poetry` (see #86). I've also manually updated the version in this PR from `0.4.0-prerelease-1` to `0.4.0-pre-1`.


pvandyken (approved):
Great, thanks!

akhanf (approved):

PR #91, opened 2021-12-07
Add release-drafter trigger

kaitj:
Sets up a separate workflow requiring a manual trigger to update the `release-drafter` changes.

This is intended to allow for a user to trigger an update to the CHANGELOG without rerunning the `bump` workflow that will cause a version bump. Also took advantage to add a conditional to the `bump` workflow so that if a PR is labelled with `skip_changelog` the workflow does not actually run and cause a version bump.

### DIFF
--- a/.github/workflows/bump_version.yml
+++ b/.github/workflows/bump_version.yml
@@ -5,22 +5,31 @@ on:
       - main
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


pvandyken (approved):

tkkuehn (approved):
LGTM!

akhanf (approved):

PR #92, opened 2021-12-08
Rename release-drafter workflow 

kaitj:
Quick patch to rename the release-drafter workflow. Missed the fact it was named the same as `bump_version` workflow. 

Also uses this PR to fix the `pr-label-checker` action that was using the wrong tag.


tkkuehn (approved):
Looks ok to me!

Issue #82, opened 2021-11-23
Function `run()` in `app.py`


pvandyken:
`run()` in `app.py` was previously used to call Snakemake in a subprocess. I've since changed that to directly call `Snakemake.main()`. The function is still used to call snakemake help, but that could also be updated to make a direct call to `main()`. Once done, is there any reason to keep the function? (Sort of an aside, but I *believe* the function's task can actually be achieved in just a few lines using `subprocess.open(...)`)

akhanf:
Yes I think once it's changed over and unused can just remove the run function.

Issue #81, opened 2021-11-23
SnakeBidsApp needs refactoring


pvandyken:
`SnakeBidsApp` in `app.py` is become quite long and convoluted. Attributes are defined across a number of functions, which rather confusing. And it breaks a few pylint rules (`too-many-instance-attributes`, `too-many-locals`) that would definitely simplify the class if fixed. So I plan on going through sometime soon and doing some refactoring.

akhanf:
I agree, thanks for taking this on!

PR #93, opened 2021-12-08
Refactor app.py

pvandyken:
Here's my promised refactor (resolving issue #81). I moved stuff around quite a bit, so sorry if the diff is a bit hard to follow.

Basically, I separated logic concerns between the argparse CLI functionality, which is now stored in `cli.py`, and the actual `Snakebids` stuff: parsing the `snakemake_dir`, reading the config file, running the app. `SnakeBidsApp` now uses the excellent [`attrs`](https://www.attrs.org/en/stable/) library for initialization.  Most of the attributes (e.g. `configfile_path`, `snakefile_path`, `config`) are determined relative to the user provided `snakemake_dir`. Previously, these were calculated within the `__init__()` function, but now are determined using `attr.Factory`, which basically lets you use a function to set a default value for an attribute. The big advantage here is that you can override any of these attributes simply by providing a value in the class constructor, which make testing *way* easier. 

All of the argparse stuff has been packaged up into standalone functions. Again, the biggest advantage here is for testing: it's much easier to test a standalone function with a nicely defined bit of logic. The `create_parser()` function returns a parser with the basic arguments. This will allow the user to import the parser and manually tweak it [we discussed this at a previous meeting: the option of defining custom args in `run.py` rather than `config.yaml`] Alternatively, the user could initialize the `SnakeBidsApp` class and then directly update the `parser` attribute it exposes. A second function: `add_dynamic_args` takes in `config.yaml` sections as variables and updates the parser with user custom args and `--filter-xx` and `--workflow-xx` magic args. Finally, `parse_snakebids_args` performs the parsing and packages everything up into the `SnakebidsArgs` class.

Another function within `cli.py`, `update_config()`, updates the config dict with info provided by the argument parser. I wasn't sure whether or not to leave that function within `app.py`. In the end, I moved it to `cli.py` but I'm happy to change if anyone objects. Its concerns bridge those of both modules. It also modifies the `config` param it receives "in-place", which can be confusing.

As mentioned in #82, I updated the Snakemake Help function to directly call `snakemake.main()`. The `run()` function in `app.py` has been removed.

A few attributes in `SnakeBidsApp` had some issues. `skip_parse_args` was broken: if a user set it to `True`, the app would have errored out. The new modes and such require argument parsing for all the new basic arguments we added. So I disabled it's functionality and added a deprecation warning. If anyone has an idea for how it could still be useful we can definitely put it back. The other attribute: `out_configfile` wasn't in the code, just the docstring. So I just deleted it.

Other than that, I tried to flesh out documentation where applicable, including more inline comments. The tests have all been updated. I didn't add any new tests, but breaking the functions into smaller bits revealed a number of things that currently have *no* tests, especially `cli.py->update_config()`, and `cli.py->add_dynamic_args()`. These should be written up at some point.

The user facing interface should be entirely the same, other than the deprecation warning (on an attribute which was broken anyway).

### DIFF
--- a/snakebids/app.py
+++ b/snakebids/app.py
@@ -76,10 +76,6 @@ class SnakeBidsApp:
     snakemake_dir : str
         Root directory of the snakebids app, containing the config file and workflow
         files.
-    skip_parse_args : bool, optional
-        @DEPRECATED (This argument currently has no functionality). If true, the
-        Snakebids app will not attempt to parse input arguments, and will only handle
-        the config file. (default: False)
     parser : ArgumentParser, optional
         Parser including only the arguments specific to this Snakebids app, as specified
         in the config file. By default, it will use `create_parser()` from `cli.py`
@@ -119,13 +115,6 @@ class SnakeBidsApp:
         Workflow snakefile will read snakebids config, create inputs_config,
         and read that in.
         """
-
-        # Issue deprecation warning for skip_parse_args
-        if self.skip_parse_args:
-            logger.warning(
-                "[DEPRECATION WARNING]: skip_parse_args will not have any effect on"
-                "this app. It will be removed in a future version"
-            )
 
         # If no SnakebidsArgs were provided on class instantiation, we compute args
         # using the provided parser
--- a/snakebids/app.py
+++ b/snakebids/app.py
@@ -2,16 +2,21 @@
 
 import argparse
 import logging
-import os
-import pathlib
-import re
-import subprocess
 import sys
+from pathlib import Path
+from typing import Any, Dict, List, Optional
 
+import attr
 import snakemake
 from colorama import Fore
 from snakemake.io import load_configfile
 
+from snakebids.cli import (
+    SnakebidsArgs,
+    add_dynamic_args,
+    create_parser,
+    parse_snakebids_args,
+)
 from snakebids.exceptions import ConfigError, RunError
 from snakebids.utils.output import (
     prepare_output,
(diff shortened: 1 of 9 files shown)


snakebids/app.py
```diff
@@ -136,317 +54,70 @@ def resolve_path(path_candidate):
 ]
 
 
-# pylint: disable=too-few-public-methods, too-many-instance-attributes
-class SnakeBidsApp:
-    """Snakebids app with config and arguments.
+def to_resolved_path(path: Union[str, os.PathLike]):
```
tkkuehn, line 57: This seems like a small enough function that it may as well be a lambda in the one place it's used (as an `attrs` converter).

tkkuehn (changes requested):
Thanks for taking on this big job! I hadn't used `attrs` before so this was a good chance to see it in action. I didn't get a chance to take a good look at the tests, but I only have one nitpick on the main body of code.

snakebids/app.py
pvandyken: Ah, yes, I had done a function as an experiment to get some typing working with Pylance (a vscode extension), but it didn't work and I never changed it back. Will adjust

pvandyken:
I should note that `pylint` doesn't apparently doesn't play well with `attr.Factory` (especially `Dict` bracket syntax `[key]`, so I had to disable a bunch of checks on the relevant classes. For me, Pylance (Pyright more specifically) fills the gap through static type checking, though I release that it's not part of our CI. But it's a pretty small gap in `pylint`'s coverage, overall, so I don't think it's a huge deal.

akhanf:
Looking great! 
I'm just going to run a quick test for one of the apps that uses skip_parse_args (to get snakemake parser for readthedocs) to see if anything is broken and make a fix if needed --

pvandyken:
Could you remind me why the readthedocs snakemake parser thing needed to use `skip_parse_args`?

akhanf:
I can't recall exactly, but v0.3.16 does fail if the option isnt provided,
but it doesn't seem to be an issue for the latest version..

akhanf (approved):
oops forgot to hit approve on this earlier, but looks great to me!

kaitj (approved):

pvandyken:
Thanks everyone. I'm REBASING and SQUASHING this PR, so be sure to pull the changes directly off the main branch once it's merged

PR #99, opened 2021-12-15
Include v0.1.0 case for release-drafter

kaitj:
Adding this into the workflow for completeness. Do we have a template that we draw from or are existing workflows currently just copying it from this repository?

Anyways, this just adds in the case for a repo to initialize release drafter if no previous tag is found (i.e. `v0.1.0`). It checks to see if `release-drafter` found a previous version, and if not, "manually" sets it to 0.1.0. The draft changelogs will still have the "Draft" title until published. 

On the deploy side, a similar check is performed, and when it reaches the publish step, it forces the tag and name of the release-drafter (this will match the `$RESOLVED_VERSION`)

It also removes the label checker, which wasn't working as intended.

Will make a separate PR to the `funcmasker-flex` repo with these changes.

### DIFF
--- a/.github/workflows/bump_version.yml
+++ b/.github/workflows/bump_version.yml
@@ -9,27 +9,17 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       
-      - name: Check PR label
-        uses: danielchabr/pr-labels-checker@v3.1
-        id: check-label
-        with:
-          hasNone: skip_changelog
-          githubToken: ${{ secrets.GITHUB_TOKEN }}
-
       - name: Update changelog
-        if: ${{ steps.check-label.outputs.passed }} == true 
         uses: release-drafter/release-drafter@v5
         id: release-drafter
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Get previous release version
-        if: ${{ steps.check-label.outputs.passed }} == true 
         run: |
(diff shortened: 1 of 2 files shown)


pvandyken:
> Do we have a template that we draw from or are existing workflows currently just copying it from this repository?

Are gists useful for this sort of thing (holding a template workflow)? I've never made one before.

pvandyken (approved):

kaitj:
> Are gists useful for this sort of thing (holding a template workflow)? I've never made one before.

Hmm, that's a good question. I would think yes with the caveat that there are also specifics of this workflow to the files in the repository. For example, the `name` and `tag` are `0.x.x` and `v0.x.x` respectively, where as in AFIDs where we have a similar workflow, the `name` and `tag` are both `v0.x.x`. 

So long as any repo using these workflows follow this current structure (and they are I believe), this isn't an issue, but it may be an issue if porting to an already existing workflow.

akhanf (approved):

tkkuehn (approved):

kaitj:
I've added the files to a gist associated with my account here: https://gist.github.com/kaitj/91ca378af402c56792af5c897a7395d9 and of course they're still available in this repo. 

I renamed the files such that `-` is suppose to match the different directories as `/` are not allowed in gist file names and moved `breaking` back under major in the gist. 

Gists are forkable, but I haven't been able to figure out how to fork it to the `khanlab` organization or how to create it directly to the organizational gist.

PR #102, opened 2021-12-18
adds get_source_path function to utils

### LINKED ISSUE #101: Snakemake_dir not updated when running app in workflow mode
`config["snakemake_dir"]` is automatically set to whatever path is passed into `SnakeBidsApp` by the user. The presents a problem in workflow mode, as `config["snakemake_dir"]` is (I think) typically used to get properly resolved paths to the Snakemake app directory. Since the Snakemake app is being replicated in workflow mode, the path in `"snakemake_dir"` should be updated to the child app. Otherwise the child maintains an unhealthy dependence on its parent.

akhanf:
Provides an alternative to snakemake's workflow.source_path(), that does
not cache the file, and can be used on directories too.

Doesn't fix #101, but provides an alternative to using `config['snakemake_dir']`

### DIFF
--- a/snakebids/utils/__init__.py
+++ b/snakebids/utils/__init__.py
@@ -1,5 +1,6 @@
 from snakebids.utils.output import (
     Mode,
+    get_source_path,
     get_time_hash,
     prepare_output,
     retrofit_output,
@@ -14,6 +15,7 @@ from snakebids.utils.snakemake_io import (
 
 __all__ = [
     "Mode",
+    "get_source_path",
     "get_time_hash",
     "glob_wildcards",
     "prepare_output",
--- a/snakebids/utils/output.py
+++ b/snakebids/utils/output.py
@@ -1,6 +1,7 @@
 """Tools to manage generation and interconversion of bidsapps and snakemake outputs."""
 
 import hashlib
+import inspect
 import itertools as it
 import json
 import os
@@ -450,3 +451,28 @@ def write_config_file(config_file: Path, data: dict, force_overwrite: bool = Fal
         yaml.add_representer(WindowsPath, path2str)
 
         yaml.dump(data, f, default_flow_style=False, sort_keys=False)
+
+
+def get_source_path(rel_path):
+    """Gets path relative to the file that calls this function.
+
+    Adapted from snakemake's workflow.source_path(), but does not do any caching.
+    Used to resolves paths for resources relative to the workflow.
+
+    Parameters
+    ----------
+    rel_path : str
+        Path to retrieve assumed to be relative to the location of
+        the calling file.
+    
+    Returns
+    -------
+    str
+        Absolute path to the file
+
+    """
+
+    frame = inspect.currentframe().f_back
+    calling_file = frame.f_code.co_filename
+    calling_dir = os.path.dirname(calling_file)
+    return Path(os.path.join(calling_dir, rel_path)).resolve()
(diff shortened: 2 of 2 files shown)


pvandyken:
I literally just ran into this problem yesterday (needing a directory reference), so I'm glad to see my wishes so immediately coalesce into a PR without any effort on my part. Thanks for pulling this together

pvandyken:
I'm kind of thinking a nicer interface may be to have a 0-argument function that just returns the `Path` to snakemake app dir or the `Snakefile` dir. Then you could have something like `snakemake_dir()/'resources/whatever'`. To me that feels more intuitive and simpler. Or perhaps you just make the argument optional and set it by default to `Path.cwd().resolve()`, which would have the same effect. That's probably the best actually

akhanf:
Ya I was also initially considering a 0-arg function, but actually the optional argument would be the best of both worlds.. wondering if the name of the function could be more meaningful then? feel free to make any edits!

pvandyken:
Would `snakemake_dir()` be good?

Also, would we keep `"snakemake_dir"` in the config? It would be more for provenance I guess?

akhanf:
The way it is set-up now, the paths are relative to the folder where the snakefile or rule file is calling it from. That seems intuitive since it's the convention used for other things, like the `script:` directive.  
We could just perhaps add both kinds of functions, `source_path`, and `snakemake_dir` as functions in snakebids utils?

tkkuehn:
Having two separate functions seems reasonable to me. I'd suggest names like `snakefile_dir` and `relative_to_snakefile` for clarity (although `source_path` does have the benefit of being similar to the Snakemake name, so I'm fine with that too).

pvandyken:
Just for clarity, @akhanf , when you say two functions, are you thinking one function relative to the snakefile and one function relative to the app directory? Or are you thinking one 0-arg function that returns the snakefile argument and on function of the form `function(path) -> path_resolved_to_snakefile`

akhanf:
The issue is for this function (and similarly for existing workflow.source_path function), the path is not relative to the snakefile or app directory, it is relative to the file the function is called from. I actually don't know how to automatically get the app directly in this way, since it could be called from a Snakefile in the root dir, a Snakefile in the workflow folder, or a rules file in any rules subdirectory (and so on)..

So I was suggesting one function that does just that (similar semantically to workflow.source_path), and then if there is a way to automatically get the snakemake working directory, another function for that. I can't think of how to implement that right now but I assume there is some way to get that? If there isn't a way to do that, then the one function would suffice. 

Both of them can be 0-arg functions for ease of use.

pvandyken:
So I just did some digging into the snakemake source. Turns out there's two properties that basically achieve our purpose:

- `workflow.basedir` is the directory containing the root level Snakefile
- `workflow.current_basedir` is the directory containing the file requesting that property (e.g. /path/to/workflow/rules)

Because snakemake uses `os.chdir()` to change the working directory when the `--directory` flag is used, I don't *believe* it's possible to recover the app base directory. It's not saved anywhere so far as I can tell: snakemake only tracks the `Snakefile` and configfile directories.

At any rate, if the above properties work, that should eliminate the need to roll out our own solution.

akhanf:
Ah that's great! Yes I agree no need to roll out some thing custom then!
Perhaps we can just document it somewhere and/or include the use of them in any examples/templates we have

Issue #105, opened 2021-12-21
Missing-argument help-message has duplicate names


akhanf:
Just made one argument in hippunfold a required one, e.g. snakebids.yml excerpt here:

```
  --modality:
    help: 'Type of image to run hippunfold on. Modality prefixed with seg will import an existing (manual) hippocampal tissue segmentation from that space, instead of running neural network (default: %(default)s)'
    required: True
    choices:
      - T1w
      - T2w
      - hippb500
      - segT1w
      - segT2w
      - cropseg
```

but when a user doesn't provide the required arg we get the following error message:

```
hippunfold: error: the following arguments are required: --modality/--modality/--modality
```
The problem here is --modality is repeated three times identically.. 

Pretty minor problem, since it's just in the help output, but something to make note of regardless. I am guessing the code that is adding hyphen alternatives might be doing this, but haven't checked it out.

pvandyken:
I believe that is likely the problem. I'll take a look

PR #106, opened 2021-12-22
Use set() to remove duplicate arg labels

pvandyken:
I believe #105 was a simple as using set() for de-duplication. #akhanf, could you do me a favor and test this PR? I don't have test cases prepared, and if I made them now, they'd have to immediately be changed to accomadate the refactoring in #93.

Addresses #105

### DIFF
--- a/snakebids/app.py
+++ b/snakebids/app.py
@@ -329,7 +329,7 @@ class SnakeBidsApp:
                 )
             else:
                 names = (name,)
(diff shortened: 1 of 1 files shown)


akhanf:
Sure, I can test it out in the branch of hippunfold I was working on

tkkuehn (approved):
LGTM!

akhanf (approved):
just tested and it is fixed, thanks!

Issue #100, opened 2021-12-15
`bids()` input lists and input zip lists


pvandyken:
### Discussed in https://github.com/akhanf/snakebids/discussions/98

<sup>Originally posted by **pvandyken** December 13, 2021</sup>
A while back, we discussed how the current attributes available on the dict returned by `generate_inputs()` are somewhat confusing, especially `"input_lists"` and `"input_zip_lists"`. The challenge with `"input_lists"` is that when you use it with `expand()`, it will expand to all possible combinations of bids attributes, some of which may not exist. The fastest work-around if this attribute must be used is probably to filter the `expand()` list with `Path.exists()`. 

On the other hand, `input_zip_lists` will always return files that do exist, but for it to work correctly, users must call `expand()` as `expand("/path/to/{wildcard}", zip, generate_inputs(...)["input_zip_lists"])`. The `zip` is necessary to change the default behaviour of `expand` away from `product`. 

I would propose two things to remedy this. First, add a function to the return of `generate_inputs()` called `expand()` or something similar that takes one positional argument: the path to be expanded. It will call the `snakemake.io.expand()` function using `zip` and all of the bids parameters originally passed to `generate_inputs()`. It will also have a `**kwargs` intended for any additional arguments. In this case, two `expands()` will be performed, the first using `zip` and the bids parameters, the second using `product` and the `**kwargs`. Something like:
```py
from snakemake.io import expand

def sb_expand(path, **kwargs):
    x = expand(path, zip, **generated_inputs["input_zip_lists"]
    return expand(x, **kwargs)
```
Optionally, we could also allow the user to specify `**kwargs` that were already passed to `generate_inputs()` as bids parameters. These would be used to filter the bids parameters. However, this feature would be inherently messy, and my personal preference would be for users to use the filter capabilities within `generate_inputs()`. 

[...]

PR #107, opened 2022-01-08
Refactor bids() function

pvandyken:
The process of working on #100 led to a lot of changes, not all of which are related. I thus thought it might be easier if I release the changes across a few PRs based on content theme. 

The main goal here was for bids() to return Paths rather than strings. Snakemake can consume Paths as input/outputs just fine, and Paths are just more convenient. 

Along the way, I refactored a few things:

- Form filename and folder in a "one-liner" list creation, rather than gradually appending items. This makes the code much more concise and clear.

- Add type annotations

### DIFF
--- a/snakebids/bids.py
+++ b/snakebids/bids.py
@@ -5,9 +5,8 @@ import logging
 import os
 import re
 from collections import OrderedDict
-from os.path import join
 from pathlib import Path
-from typing import Optional
+from typing import Optional, Union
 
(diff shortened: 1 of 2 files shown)


kaitj (approved):
Looks good to me.

tkkuehn (approved):
Looks good to me! Much cleaner than it used to be.

pvandyken:
@akhanf, am I good to merge this in?

akhanf (approved):

PR #111, opened 2022-01-12
Enable Snakebids apps to generate boutiques descriptors

tkkuehn:
This PR enables Snakebids apps to generate [boutiques](https://github.com/boutiques/boutiques) descriptors that specify their CLI.

This was a bit more complicated to implement end-to-end than I expected, so it changes a fair amount of the codebase in ways that are not necessarily ideal. Here's a rundown, and I welcome any feedback for alternate ways of doing this:

- To set up the CLI, I set up two [subparsers](https://github.com/boutiques/boutiques) for a Snakebids app: `run.py run` contains the standard arguments for running a Snakebids workflow as usual, and `run.py boutiques {descriptor_path}` saves a boutiques descriptor to the provided path.
- To accommodate the ability to add dynamic arguments to an app with subparsers involved, I had to change the return value of `create_parser` to a tuple with the overall parser and the `run` subparser.
- To assign both parsers as attributes of the `SnakeBidsApp` class, I had to implement a class method to instantiate the class from the filesystem (I couldn't figure out a good way of using an `attrs` factory with a function that returns multiple values), which appears to be recommended practice anyway.
- This change to the defaults in `SnakeBidsApp`'s `__init__` meant that I had to change the `run.py` in the project template, which means this change will not be backwards compatible.

Note: the generated boutiques descriptors here will not contain any Snakemake arguments at this point. Including those is a feature I'd like to have in the future, but this seemed like a good place to cut off this PR.

Note 2: I had to change the pre commit config to permit Python 3.10 (I should really have done this in a separate PR, oh well)


pvandyken:
Thanks for all the work @tkkuehn. I'm still not totally up to speed on what `boutiques` does, but from the little I know this looks like a great addition!

I'm wondering if a better way to handle the CLI is to have a `--boutiques` optional argument that takes a path as a value: `run.py --boutiques /path/to/output`. When supplied, it would ignore anything else provided and just do the boutiques stuff. This isn't a fully formulated idea yet: for instance, the `bids_dir`, `output`, etc are typically mandatory, and they shouldn't be for boutiques. So I'll think about that some more.

One thing I will say is that I had envisioned `create_parser()` as something workflow developers could import and customize. So it would be nice to keep it simple and streamlined, and find a way to bury the boutiques functionality a bit. Still thinking of how to do that though.

I'm in favour of putting the logic to initialize `SnakeBidsApp` in it's own function. As you say, it is recommended and it helps keep things flexible. One hack we could implement to prevent breaking changes is to override the `__new__()` method for cases when only a single argument (`snakemake_dir`) is provided (which should be universally the case in production code). So something like:

```py
def __run__(cls, snakemake_dir, *args, **kwargs):
    if not args and not kwargs:
        return cls.from_filesystem(snakemake_dir)
    return cls(snakemake_dir, *args, **kwargs)
```
I actually don't think this is a super hacky solution; overriding `__run__()` is common enough. On the other hand, we're still in `v0.x`, so breaking changes also aren't a huge deal.

kaitj:
Skimmed through the code this morning, and it looks good to me.

I was thinking along the same lines as @pvandyken of the use of a `--boutiques` flag.

pvandyken:
Okay, here's a more complete idea:

Don't make any changes to `create_parser()`. Instead, at the beginning of `run_snakemake()`, after `add_dynamic_args()` but before `parse_snakebids_args`, we run `sys.argv` through a boutique preprocesser. It will look for the `--boutique[-whatever]` flag along with a `Path` value. If it finds it, we skip the regular parsing of args and act accordingly. Otherwise, `run_snakemake()` can continue as normal.

I had tried to make the above scheme more generic. For instance, we may eventually want another of these sort of optional arguments that completely change `run.py` behaviour. It's hard to make it generic without any ideas of what these other optional args may be, so for now I think hardcoding the behaviour in would be fine.

tkkuehn:
I attempted to override `__new__` along the lines @pvandyken suggested, but the issue strikes that when `__new__` returns an instance of its class, Python calls `__init__` with the arguments provided to `__new__`. Because the overridden `__new__` only has one argument and defers the rest to `from_filesystem`, the `__init__` call fails with not enough arguments. I don't think it's worth messing with the class definition to fix this, so I think we should just make this change backwards-incompatible.

Here's the `__new__` redefinition I tried:

```
    def __new__(cls, snakemake_dir, *args, **kwargs):
        if not args and not kwargs:
            print("Producing from filesystem...")
            return cls.from_filesystem(snakemake_dir)
        kwargs = kwargs | {"snakemake_dir": snakemake_dir}
        return super().__new__(cls)
```

pvandyken:
Does it not work if you pass `*args` and `**kwargs` to `super().__new__`?

```py
  def __new__(cls, snakemake_dir, *args, **kwargs):
      if not args and not kwargs:
          print("Producing from filesystem...")
          return cls.from_filesystem(snakemake_dir)
      kwargs = kwargs | {"snakemake_dir": snakemake_dir}
      return super().__new__(cls, *args, **kwargs)
```

tkkuehn:
> Does it not work if you pass `*args` and `**kwargs` to `super().__new__`?

Ah, I should have mentioned that -- it results in `TypeError: object.__new__() takes exactly one argument (the type to instantiate)`

tkkuehn:
Okay, unless I'm missing something re: the `__new__` discussion above, this should be ready for another look.

pvandyken:
Was just taking a look at `__new__` to understand what I was getting wrong. I guess the issue was that `super()` was invoking the base `object` constructor, which doesn't take any additional arguments. My apologies for sending you down a rabbit trail.

Otherwise, aside from one or two comments that I'll leave, everything looks good. One more general idea that could (should) be reserved for a new PR: so long as we're changing the API anyway, it may be a good idea to create a new function called `run_snakemake` that initializes the `SnakeBidsApp` and calls `SnakeBidsApp.run_snakemake()` all in one go. So instead of needing `SnakeBidsApp.from_filesystem("path/to/snakemake").run_snakemake()`, you can just do `run_snakemake("path/to/snakemake")`. Unless the API for `SnakeBidsApp` becomes significantly expanded, I think very few people will need to modify `SnakeBidsApp` before calling `run_snakemake()`.

snakebids/app.py
```diff
@@ -95,19 +92,43 @@ class SnakeBidsApp:
...
+        ---------
+        snakemake_dir : str
+            Root directory of the snakebids app, containing the config file and workflow
+            files.
+        args : SnakebidsArgs, optional
+            Arguments to use when running the app.
+        skip_parse_args : bool
+            The SnakebidsApp won't parse arguments if true.
+        """
+        parser = create_parser()
```
pvandyken, line 119: Could `parser` be made a parameter as well, with the default as `create_parser()`? That way `parser` could be customized.

snakebids/project_template/{{cookiecutter.app_name}}/{{cookiecutter.app_name}}/run.py
```diff
@@ -6,12 +6,13 @@
 
 def get_parser():
     """Exposes parser for sphinx doc generation, cwd is the docs dir"""
-    app = SnakeBidsApp("../", skip_parse_args=True)
+    app = SnakeBidsApp.from_filesystem("../", skip_parse_args=True)
     return app.parser
```
pvandyken, line 10: I think this comment is something I missed previously, but it looks as if this function can just call `create_parser()` rather than going through `SnakeBidsApp`, but I might be mistaken (I haven't worked through the `project_template` code much)

snakebids/app.py
```diff
@@ -73,20 +70,20 @@ class SnakeBidsApp:
...
-    snakemake_dir : str
+    snakemake_dir : Path
         Root directory of the snakebids app, containing the config file and workflow
         files.
-    parser : ArgumentParser, optional
+    parser : ArgumentParser
         Parser including only the arguments specific to this Snakebids app, as specified
         in the config file. By default, it will use `create_parser()` from `cli.py`
-    configfile_path : str, optional
+    configfile_path : str
```
kaitj, line 79: Don't think the `configfile_path` is autocalculated anymore with the removal of `_get_file_paths` unless I am missing something (at least not when the class first instantiated)

snakebids/app.py
```diff
@@ -73,20 +70,20 @@ class SnakeBidsApp:
...
         Relative path to config file (relative to snakemake_dir). By default,
-        autocalculates based on snamake_dir
-    snakefile_path : str, optional
+        autocalculates based on snakemake_dir
+    snakefile_path : str
```
kaitj, line 82: Same with the `Snakefile`

kaitj (commented):
Overall looks good to me, a few minor comments.

Barring additional changes, I think this is good to merge.

kaitj:
> One more general idea that could (should) be reserved for a new PR: so long as we're changing the API anyway, it may be a good idea to create a new function called `run_snakemake` that initializes the `SnakeBidsApp` and calls `SnakeBidsApp.run_snakemake()` all in one go. So instead of needing `SnakeBidsApp.from_filesystem("path/to/snakemake").run_snakemake()`, you can just do `run_snakemake("path/to/snakemake")`. Unless the API for `SnakeBidsApp` becomes significantly expanded, I think very few people will need to modify `SnakeBidsApp` before calling `run_snakemake()`.

I can see the advantages of having a separate function, but I also like the idea of having it separate where (1) builds the app and (2) runs the app. It also seems to simplify the boutiques generator. Anyways, I think we can open up an issue / discussion thread about this and leave it for another PR.

tkkuehn:
I think adding a separate `run_snakemake` function in `app.py` that just encapsulates the `SnakeBidsApp.from_filesystem("path/to/snakemake").run_snakemake()` would be nice and wouldn't do any harm. But yeah, best to discuss in a separate issue.

tkkuehn:
Addressed the last round of comments, so I think this is good to go if one of you want to merge?

kaitj (approved):

pvandyken (approved):

akhanf:
This breaks existing apps since it makes changes to the interface for SnakeBidsApp -- was this intended?
```
(snakebids-JiNcMeMJ-py3.8) alik@AFI-CBS-H-6:/local/scratch/hippunfold$ ./hippunfold/run.py 
Traceback (most recent call last):
  File "./hippunfold/run.py", line 24, in <module>
    main()
  File "./hippunfold/run.py", line 16, in main
    app = SnakeBidsApp(
TypeError: __init__() missing 3 required positional arguments: 'parser', 'snakefile_path', and 'config'
```

tkkuehn:
> This breaks existing apps since it makes changes to the interface for SnakeBidsApp -- was this intended?

Yes -- I tried to maintain the existing interface by overriding `SnakeBidsApp`'s `__run__` method like Peter suggested, but couldn't find a way to make that work without even more changes to the class.

The solution to the break with existing apps is to update their `run.py` to be consistent with the version introduced in this PR (or not upgrade Snakebids once it's released).

With the changes to the Boutiques interface since the first draft of this PR, though, I could reverse the changes to the SnakeBidsApp interface (while maintaining the new Boutiques stuff) if we want to be sure to maintain backwards compatibility.

pvandyken:
Since the new `SnakeBidsApp` interface is not actually related to Boutiques, an immediate course of action may be to put the `SnakeBidsApp` change into a new branch, and then this could be merged without controversy. That would give use some time to discuss interface changes.

akhanf:
If the interface changes are things we are going to change anyways in another PR, then I guess I am ok with breaking backwards compatibility.. I'm not just fully up to date on what interface changes to SnakeBidsApp are actually needed fo this (e.g. I don't know why some parameters that were optional became required?)

tkkuehn:
Ok, I just stripped the SnakeBidsApp interface changes from this PR. I'll restore them in a separate branch for further discussion, but this PR should no longer contain any breaking changes.

pvandyken (approved):

akhanf:
Had a chance to go through this in more detail now and also test out the boutiques descriptors.

Re: the new snakebids CLI, all looks good except:
1. when you run `snakebids` without any arguments, instead of getting the help, you just get an error (confused me for a while!):
```
(snakebids-9xW862mn-py3.8) alik@AFI-CBS-H-6:/local/scratch/snakedwi$ snakebids 
Traceback (most recent call last):
  File "/local/scratch/pypoetry_venv/snakebids-9xW862mn-py3.8/bin/snakebids", line 5, in <module>
    main()
  File "/local/scratch/snakebids-boutiques/snakebids/admin.py", line 58, in main
    args.func(args)
AttributeError: 'Namespace' object has no attribute 'func'
```
2. would also be nice to indicate that `out_path` in snakebids boutique should be a json file.


Re: the actual generated boutiques descriptors we get, testing them with `bosh` from boutiques:
3. There are `null` values in the `description` values for the `--path-*` options, which cause the validator to fail:
```
[ ERROR ] None is not of type 'string'

Failed validating 'type' in schema['properties']['inputs']['items']['properties']['description']:
    {'description': 'Input description.',
     'id': 'http://github.com/boutiques/boutiques-schema/input/description',
     'type': 'string'}

On instance['inputs'][19]['description']:
    None
```
Perhaps this is ok though, since there likely will be other type fixes one would need to make manually (e.g. I also had to change `String` to `Number` for some optional inputs.


4. We should add some of the snakemake arguments to the descriptor too (dryrun, and cores for sure)


Of the above points, I think would be nice to have 1. for this PR, but other points we can deal with in future issues/PRs.. Likely more things will come up as we explore boutiques a bit more!

tkkuehn:
>     1. when you run `snakebids` without any arguments, instead of getting the help, you just get an error (confused me for a while!):

I _think_ this should be fixed now -- it stemmed from a bug in argparse in Python 3.7 and 3.8. That said, I wasn't able to reproduce this exact bug using Python 3.8 on the CBS server (I got a slightly different error message), so do you mind giving it another shot @akhanf ?

I added a bit more help text to cover pont 2, and agree that we should address points 3 and 4 in a separate PR.

akhanf (approved):
looks good to me!

PR #115, opened 2022-01-14
Refactor bids.py

pvandyken:
Step 2 in the road toward #100!

This one is mostly just moving code around. I split `bids.py` into three separate files: `bids.py` containing `bids()` and `print_boilerplate()`, `filtering.py` containing the `filter_list()` function family, and `input_generation.py` containing everything to do with `generate_inputs()`. This split will become especially important for upcoming PRs, where the file sizes will increase. Note that I didn't change any code, just moved stuff around.

read_bids_tags has been moved to a new `utils.py` file. There's a few more functions coming down the pipeline that will also go into this file. I slightly modified how the function worked: it was previously finding the `bids_tags.json` via relative path. It now uses `importlib`, which is able to find the resources relative to the root `snakebids` package.

Aside from that, none of the code has actually changed.


pvandyken:
Also, not sure if either `more-itertools` or `hypothesis` is actually used in this particular PR, but they'll be needed later.

tkkuehn (approved):
Looks good to me!

kaitj (approved):
Also lgtm!

pvandyken:
Great, I'm going to go ahead and merge this then.

PR #116, opened 2022-01-17
Improve filter_list

pvandyken:
The zip list filter previously could only filter one entity-value at a time. In other words, you could search for a single subject, a single aquisition, etc, but not subjects 4 and 5 in the AP direction. This PR updates it to support any arbitrary combination of entities. It also supports regex_search, using the same interface as `pybids.get()`. A formal test was created by copying out the doc string (which admittedly doesn't add much but it's easier for me to run the test in VS-code that way). I added a new condition to the formal test to test searching by entity combinations.

Due to my use of `zip()` within `filter_list`, each `zip_list` is now returned as a tuple instead of a list. I didn't feel the need to correct this: realistically the user should not be `append()`ing anything to the `zip_lists`, and if they really want to, they can convert to a list themselves. I have not yet gone through and changed every function creating a `zip_list` so that they all make tuples instead, but this will be done an upcoming PR (fairly soon...). In the meantime, tuples and lists are similar enough that I don't think it will make a difference.

### DIFF
--- a/snakebids/core/filtering.py
+++ b/snakebids/core/filtering.py
@@ -1,4 +1,40 @@
-def filter_list(zip_list, wildcards, return_indices_only=False):
+import itertools as it
+import operator as op
+import re
+from typing import Dict, List, Tuple, Union, overload
+
+import more_itertools as itx
+from typing_extensions import Literal
+
+from snakebids.utils.utils import matches_any
+
+
+@overload
+def filter_list(
+    zip_list,
+    filters: Dict[str, Union[str, List[str]]],
+    return_indices_only: Literal[False] = ...,
+    regex_search: bool = ...,
+) -> Dict[str, Tuple[str]]:
+    ...
+
+
+@overload
+def filter_list(
+    zip_list,
+    filters: Dict[str, Union[str, List[str]]],
+    return_indices_only: Literal[True] = ...,
+    regex_search: bool = ...,
+) -> Tuple[int]:
+    ...
+
+
+def filter_list(
+    zip_list: Dict[str, List[str]],
+    filters: Dict[str, Union[str, List[str]]],
+    return_indices_only=False,
+    regex_search=False,
+):
     """This function is used when you are expanding over some subset of the
     wildcards i.e. if your output file doesn't contain all the wildcards in
     input_wildcards
@@ -8,11 +44,14 @@ def filter_list(zip_list, wildcards, return_indices_only=False):
     zip_list : dict
         generated zip lists dict from config file to filter
 
-    wildcards : dict
+    filters : dict
         wildcard values to filter the zip lists
 
     return_indices_only : bool, default=False
         return the indices of the matching wildcards
+
+    regex_search : bool, default=False
+        Use regex matching to filter instead of the default equality check.
 
     Returns
     -------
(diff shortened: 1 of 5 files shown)


kaitj (commented):
First pass going through the code, the PR looks good to me. 

Out of curiosity, was wondering what the use case was for the emails - noticed there is a test case testing for that.

kaitj (approved):

akhanf (approved):
Looks good to me!
Thanks!

akhanf:
actually before you merge this, let me just do a couple quick dry-run tests on some workflows that use ziplists to confirm nothing breaks

tkkuehn (approved):
Looks good to me!

pvandyken:
Great, I'll merge this as soon as @akhanf reports back

akhanf:
just checking now -- something is getting broken in snakedwi  with `snakebids.get_filtered_ziplist_index` .. looking into it now

akhanf:
ah ok, instead of `get_filtered_ziplist_index()` I really should just be using `filter_list` with `return_indices=True`.

I am thinking now we can also eliminate the `get_filtered_ziplist_index` function with this refactoring too?

akhanf (approved):
I will re-approve this now! Will leave it up to you to remove `get_filtered_ziplist_index` now or later.. I think the only place it was ever used is where I just replaced it in snakedwi..

akhanf:
for posterity, it actually wasn't get_filtered_ziplist_index that was broken, the problem was just with the move to tuples, but I've dealt with that now in the workflow..

pvandyken:
Again I'm making presumptive type changes that are causing unexpected problems! Are you sure moving to tuples won't be an issue @akhanf? Now's the time to change it.

I admit I'm a little confused with why `get_filtered_ziplist_index` is suddenly obsolete, as it doesn't do the same thing as `filter_list`. Admittedly, it's purpose seems niche: I can't imagine where you would use it. But if you say we don't need it, I'll be rid of it.

akhanf:
Haha no worries.. My last message was trying to explain that it wasn't actually anything about get_filtered_ziplist_index becoming obsolete, I was mistaken for a moment.. it was just the move to tuples, but all good now

pvandyken:
Okay, so just to be clear, do I remove `get_filtered_ziplist_index` or no?

akhanf:
I see it as redundant, so can remove it

pvandyken:
Sorry for completely resurrecting this PR, but I took a look at it again and found that all was not right.

I ran a bunch of benchmarks, and as it turns out, all I managed to do with my first revision was make the function take twice as long to complete:

```
100,000 repeats:
New version: 1.5346982000628486
Old version (corrected for new features): 0.7702010000357404
```
(In the above test, the old version was updated to be fully featured as the new version)

So I felt bad about that and made some attempts to correct the situation. I was actually unable to write anything consistently faster than the original version, but came up with one using set intersections. It's still no faster than the original, but it's really close:
```
500,000 repeats:
New version: 4.596991300000809
Old version (corrected for new features): 4.5636120000854135
```

Along the way, I did discover an important bug with the old version: If you had two filters, say `{"subject": "03", "dir": "AP"}`, and there was no subject "03" but there was an AP direction, you'd expect no results to be returned. Instead, the function was returning all the entities with dir="AP". So that's been corrected.

At any rate, the use of `set.intersection` is largely a matter of style. I can go back to the old version and port all the improvements if people prefer.

Also, `filter_list` is now returning `lists` instead of `tuples` again. Hopefully this doesn't cause more problems again @akhanf! Of course, if people like the idea of tuples, we can still do that.

pvandyken:
For reference, the ported old version would look like this:
```python
def filter_list(...):
    if regex_search:
        match_func = re.match
    else:
        match_func = op.eq

    keep_indices = [*range(len(zip_list[next(iter(zip_list))]))]
    for key, val in wildcards.items():
        # get indices where wildcard exists
        if key not in zip_list.keys():
            continue
        indices = [
            i for i, v in enumerate(zip_list[key]) if matches_any(
                v, itx.always_iterable(val), match_func
            )
        ]
        keep_indices = [x for x in indices if x in keep_indices]
    ...
```
I have no issues using this if people prefer. The `set.intersection` approach is the result of a somewhat self-fulfilling experiment to see if I could make anything faster.

pvandyken:
I broke down the crazy `{*range(len(....))}` line and put it in a separate function for readability's sake.

tkkuehn (approved):

PR #118, opened 2022-01-25
moved project_template to package folder

### LINKED ISSUE #117: project_template not included in pypi build
Since we moved `project_template` to the root level it is no longer included in the packaging, thus snakebids-create is broken. Note this is the same issue under discussions here https://github.com/akhanf/snakebids/discussions/103

I think it is best to move this back into the `snakebids` folder to get snakebids-create working again.

akhanf:
closes #117


pvandyken:
Yep, makes sense

pvandyken (approved):

Issue #119, opened 2022-01-26
bids() returning Path instead of str breaks workflows


akhanf:
It looks like returning `Path` objects instead of `str` can break things in some workflows.

e.g. using snakemake's `multiext` with `bids()` as [here](https://github.com/akhanf/snakedwi/blob/eb124e6784ff08ca5190efc5aed185d9397903b9/snakedwi/workflow/rules/prepdwi.smk#L12) throws an error because it tries to join a Path and a str with `+`.
```
Traceback (most recent call last):
  File "/local/scratch/pypoetry_venv/snakebids-JiNcMeMJ-py3.8/lib/python3.8/site-packages/snakemake/__init__.py", line 593, in snakemake
    workflow.include(
  File "/local/scratch/pypoetry_venv/snakebids-JiNcMeMJ-py3.8/lib/python3.8/site-packages/snakemake/workflow.py", line 1186, in include
    exec(compile(code, snakefile.get_path_or_uri(), "exec"), self.globals)
  File "/local/scratch/snakedwi/snakedwi/workflow/Snakefile", line 65, in <module>
  File "/local/scratch/pypoetry_venv/snakebids-JiNcMeMJ-py3.8/lib/python3.8/site-packages/snakemake/workflow.py", line 1186, in include
    exec(compile(code, snakefile.get_path_or_uri(), "exec"), self.globals)
  File "/local/scratch/snakedwi/snakedwi/workflow/rules/prepdwi.smk", line 16, in <module>
    shell('cp -v {in_file} {out_file}')
  File "/local/scratch/pypoetry_venv/snakebids-JiNcMeMJ-py3.8/lib/python3.8/site-packages/snakemake/io.py", line 1192, in multiext
    return [flag(prefix + ext, "multiext", flag_value=prefix) for ext in extensions]
  File "/local/scratch/pypoetry_venv/snakebids-JiNcMeMJ-py3.8/lib/python3.8/site-packages/snakemake/io.py", line 1192, in <listcomp>
    return [flag(prefix + ext, "multiext", flag_value=prefix) for ext in extensions]
TypeError: unsupported operand type(s) for +: 'PosixPath' and 'str'
```

[...]

PR #120, opened 2022-01-26
Bids returns str instead of Path

pvandyken:
Hotfix for #119

Returning a Path was causing all sorts of unpredictable errors within Snakemake workflows, apparently due to lack of full Path compatibility in Snakemake


akhanf (approved):
looks good, same fix I was making locally

PR #121, opened 2022-01-26
Fix relative path bug in configfile

### LINKED ISSUE #113: Config file inside source workflow is being modified during execution, even with bids-app mode
Not sure when this was introduced, but this is causing a major problem for containerized deployment (eg of hippunfold) since the config file in the installed workflow is read-only. Just realized it is happening, so haven't had a chance to dive deeper, but I am hoping to resolve soon.

akhanf:
Closes #113 

Fixes bug where configfile was being incorrectly updated, was occurring because configfile was incorrectly being made an absolute path.

### DIFF
--- a/snakebids/app.py
+++ b/snakebids/app.py
@@ -52,30 +52,21 @@ CONFIGFILE_CHOICES = [
 ]
 
 
-def _get_configfile_paths(choices: List[str], file_name: str):
+def _get_file_paths(choices: List[str], file_name: str):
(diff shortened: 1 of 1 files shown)


kaitj (approved):
This one looks good to me. 

Would propose merging this in prior to merging of #111 (boutiques) since some things conflicts may need to be resolved and there may be some future updates? Plus it seems like there might be some further discussion regarding that PR.

PR #125, opened 2022-01-29
Rename bids.py

pvandyken:
I've been getting frustrated a few times now with not being able to import `bids.BIDSLayout` due to the current name of `snakebids.core.bids`. So this just changes the name to something unique. Makes no difference to imports: `bids` can still be imported directly from `snakebids`.


kaitj:
From a maintenance perspective, renaming seems fine, just wondering if you'd be able to provide an example where `bids.BIDSLayout` couldn't be imported due to a naming conflict?

akhanf (approved):

pvandyken:
It would happen during development any time I set up a `if __name__ == "__main__":` block at the end of any of the `snakebids.core.x` files for a quick test of something and I ran the file directly. I don't think it ever would have affected the end user.

PR #128, opened 2022-02-04
Update workflows for branch protection

kaitj:
This PR updates the workflows so protected branches can be used: 
- Switches push action to use `CasperWA/push-protected@v2` with PAT token
- Changes `bump_version.yml` trigger from `push` to `pull_request`. Only runs if this is triggered due a merged PR. Functionally the same as before.
- Added specific branches to `test.yml` to disable duplicate testing triggered by protected push.
  - Runs on all branches except those that start with `push-action/`, which is a branch created by the `CasperWA/push-protect` action.

@akhanf This should allow for branch protections to be re-enabled! Will need to generate a personal access token (saved as `BP_PAT_TOKEN` in the repository secrets) with repo access for the `push-protect` to work.

### DIFF
--- a/.github/workflows/bump_version.yml
+++ b/.github/workflows/bump_version.yml
@@ -1,13 +1,16 @@
 name: Bump version
 on:
-  push:
-    branches:
-      - main
+  pull_request:
+    types: [closed]
 jobs:
   build:
+    if: github.event.pull_request.merged == true 
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


pvandyken (approved):

tkkuehn (approved):
Looks good to me, as long as I'm remembering correctly that you tested this on your fork and it worked as expected

akhanf (approved):
looks good to me, I added the BP_PAT_TOKEN so should be good to go! I'll merge this in, then turn on branch protection

PR #132, opened 2022-02-07
Fix auto-assign

### LINKED ISSUE #131: Actions auto-assign
Just noticed from #130 that the `auto-assign` workflow isn't working and is running into a similar issue that has been encountered on other projects. Not a pressing issue and I believe is a wider github actions issue. The problem arises whenever you have an action that needs permissions on a forked repo instead of another branch within the main repo.

kaitj:
Resolves #131. 

Actions that require write permissions to the repo (e.g. committing and pushing during the action) or repo secrets (e.g. the use of `secrets.GITHUB_TOKEN`) does not work on forks. When `pull_request` is the trigger, the workflow only has **read access** and **if user (without write access)** authored a PR from a fork, any workflows requiring repository secrets or write access will fail. The workaround is to use `pull_request_target`, which functionally is the same as `pull_request`, but also enables **all** users to have **write access** and access to repository secrets through the workflow. 

The changes from this PR:
* Splits up reviewer assignment from test workflow for security purposes
  * `pull_request_target` has some security vulnerabilities when paired with the `actions/checkout` using default parameters. To get around this, i moved the reviewer assignment to its own workflow that does not require the commit to be checked out and paired with `pull_request_target`, keeping the test workflow the same. This does mean all PRs will be assigned to reviewers prior to the completion of unit testing.

_Here's a long [video](https://www.youtube.com/watch?v=Ww5gjAm-2GY&t=1946s) that goes into a bit of detail about github actions and pull requests._

### DIFF
--- a/.github/workflows/assign.yml
+++ b/.github/workflows/assign.yml
@@ -1,0 +1,14 @@
+name: Assign reviewer 
+
+on:
+  pull_request_target: 
+
+jobs: 
+  assign:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Assign reviewer
+        uses: kentaro-m/auto-assign-action@v1.1.2
+        with:
+          repo-token: "${{ secrets.GITHUB_TOKEN }}" 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -118,7 +118,7 @@ jobs:
     runs-on: ubuntu-latest
     name: Reviewer assignment
     needs: [test]
-    if: github.event_name == 'pull_request' && github.event.pull_request.assignee == null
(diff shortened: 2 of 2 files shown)


kaitj:
Tried to quickly fix it by replacing pull_request with pull_request_target, but it seems to just skip the reviewer assignment now. When this PR is ready, I'll mark it `ready for review`

kaitj:
Should be ready for review now. 

Tested it by creating a PR from @tkkuehn's fork to my own fork [here](https://github.com/kaitj/snakebids/pull/20). Note that the reviewer can't actually get assigned in this test as none of the other listed users are collaborators on my fork. Workflow was still able to run and assign an author however.

tkkuehn (approved):
Seems reasonable to me!

akhanf (approved):

PR #133, opened 2022-02-22
Add ability to use pybids db

kaitj:
This PR adds in functionality to save the `pybids` db. 

It adds a `pybids_db` to the config, where it stores the db directory, and a flag as to whether the db should be overwritten. Wanted this to be separate from the `pybids_inputs`. This also updates the `generate_input.py` to make use of these config options. If the directory is left blank (e.g. `''`), the database will be set to `None` (if left as `''`, it saves the db in the current working directory). Also added in a few small test cases to make sure the db is working as expected.

~@pvandyken I didn't end up using the `pytest.fixtures` and `tmpdir` (mentioned last week) to create the new temporary data during testing - looked as if these temp files would be created at the beginning of the test which would have been picked up by `pybids` and wanted to test for changes.~

### DIFF
--- a/snakebids/core/input_generation.py
+++ b/snakebids/core/input_generation.py
@@ -1,7 +1,7 @@
 import json
 import logging
-import os
 import re
+from pathlib import Path
 from typing import Dict, List, Tuple
 
 from bids import BIDSLayout, BIDSLayoutIndexer
@@ -12,10 +12,12 @@ from snakebids.utils.utils import read_bids_tags
 _logger = logging.getLogger(__name__)
 
 
-# pylint: disable=too-many-arguments
+# pylint: disable=too-many-arguments,too-many-locals
 def generate_inputs(
     bids_dir,
     pybids_inputs,
+    pybids_database_dir=None,
+    pybids_reset_database=False,
     derivatives=False,
     pybids_config=None,
     search_terms=None,
@@ -164,23 +166,14 @@ def generate_inputs(
 
     search_terms = _generate_search_terms(participant_label, exclude_participant_label)
 
-    if os.path.exists(bids_dir):
-        # generate inputs based on config
-        layout = BIDSLayout(
-            bids_dir,
-            derivatives=derivatives,
-            validate=False,
-            # In the next version of pybids, config will accept Paths, so we won't
-            # need this long stringify line
-            config=str(pybids_config) if pybids_config is not None else pybids_config,
-            indexer=BIDSLayoutIndexer(validate=False, index_metadata=False),
-        )
-    else:
-        _logger.info(
-            "bids_dir does not exist, skipping PyBIDS and using "
-            "custom file paths only"
-        )
-        layout = None
+    # Generates a BIDSLayout
+    layout = _gen_bids_layout(
+        bids_dir=bids_dir,
+        derivatives=derivatives,
+        pybids_config=pybids_config,
+        pybids_database_dir=pybids_database_dir,
+        pybids_reset_database=pybids_reset_database,
+    )
 
     # this will populate input_path, input_lists, input_zip_lists, and
     # input_wildcards
@@ -236,6 +229,54 @@ def generate_inputs(
         }
 
     return inputs_config_dict
+
+
+def _gen_bids_layout(
+    bids_dir,
+    derivatives,
+    pybids_database_dir,
+    pybids_reset_database,
+    pybids_config=None,
+):
+    """Create (or reindex) the BIDSLayout if one doesn't exist,
+    which is only saved if a database directory path is provided
+    """
+
+    # Set db dir to None (otherwise saves to parent dir)
+    if Path(bids_dir):
+        # Check for database_dir
+        # If blank, assume db not to be used
+        if pybids_database_dir == "":
+            pybids_database_dir = None
+        # Otherwise check for relative path and update
+        elif (
+            pybids_database_dir is not None
+            and not Path(pybids_database_dir).is_absolute()
+        ):
+            pybids_database_dir = None
+            _logger.warning(
+                "Absolute path must be provided, database will " "not be used"
+            )
+
+        layout = BIDSLayout(
+            bids_dir,
+            derivatives=derivatives,
+            validate=False,
+            # In the next version of pybids, config will accept Paths, so we won't
+            # need this long stringify line
+            config=str(pybids_config) if pybids_config is not None else pybids_config,
+            database_path=pybids_database_dir,
+            reset_database=pybids_reset_database,
+            indexer=BIDSLayoutIndexer(validate=False, index_metadata=False),
+        )
+    else:
+        _logger.info(
+            "bids_dir does not exist, skipping PyBIDS and using "
+            "custom file paths only"
+        )
+        layout = None
+
+    return layout
 
 
 def write_derivative_json(snakemake, **kwargs):
--- a/snakebids/project_template/{{cookiecutter.app_name}}/config/snakebids.yml
+++ b/snakebids/project_template/{{cookiecutter.app_name}}/config/snakebids.yml
@@ -6,10 +6,10 @@ debug: False
 
 derivatives: False #will search in bids/derivatives if True; can also be path(s) to derivatives datasets
 
-#list of analysis levels in the bids app 
+#list of analysis levels in the bids app
 analysis_levels: &analysis_levels
  - participant
-  
+
 
 #mapping from analysis_level to set of target rules or files
 targets_by_analysis_level:
@@ -33,41 +33,46 @@ pybids_inputs:
       - acquisition
       - task
       - run
-   
+
+#this configures the options to save the BIDSLayout
+# by default, database is not saved (uncomment to save)
+# NOTE: pybids_db_dir must be an absolute path
+# pybids_db_dir: '/path/to/db_dir' # Leave blank if you do not wish to use this
+# pybids_db_reset: False # Change this to true to update the database
 
 #configuration for the command-line parameters to make available
 # passed on the argparse add_argument()
 parse_args:
 
-#---  core BIDS-app options --- (do not modify below) 
+#---  core BIDS-app options --- (do not modify below)
 
   bids_dir:
-    help: The directory with the input dataset formatted according 
+    help: The directory with the input dataset formatted according
           to the BIDS standard.
 
   output_dir:
-    help: The directory where the output files 
+    help: The directory where the output files
           should be stored. If you are running group level analysis
           this folder should be prepopulated with the results of the
           participant level analysis.
 
-  analysis_level: 
-    help: Level of the analysis that will be performed. 
+  analysis_level:
+    help: Level of the analysis that will be performed.
     choices: *analysis_levels
 
   --participant_label:
-    help: The label(s) of the participant(s) that should be analyzed. The label 
-          corresponds to sub-<participant_label> from the BIDS spec 
-          (so it does not include "sub-"). If this parameter is not 
-          provided all subjects should be analyzed. Multiple 
+    help: The label(s) of the participant(s) that should be analyzed. The label
+          corresponds to sub-<participant_label> from the BIDS spec
+          (so it does not include "sub-"). If this parameter is not
+          provided all subjects should be analyzed. Multiple
           participants can be specified with a space separated list.
     nargs: '+'
 
   --exclude_participant_label:
-    help: The label(s) of the participant(s) that should be excluded. The label 
-          corresponds to sub-<participant_label> from the BIDS spec 
-          (so it does not include "sub-"). If this parameter is not 
-          provided all subjects should be analyzed. Multiple 
+    help: The label(s) of the participant(s) that should be excluded. The label
+          corresponds to sub-<participant_label> from the BIDS spec
+          (so it does not include "sub-"). If this parameter is not
+          provided all subjects should be analyzed. Multiple
           participants can be specified with a space separated list.
     nargs: '+'
 
@@ -89,7 +94,3 @@ parse_args:
 #singularity containers
 singularity:
     fsl: 'docker://brainlife/fsl/6.0.0'
-
-
-
-
--- a/snakebids/project_template/{{cookiecutter.app_name}}/workflow/Snakefile
+++ b/snakebids/project_template/{{cookiecutter.app_name}}/workflow/Snakefile
@@ -10,6 +10,8 @@ config.update(
     snakebids.generate_inputs(
         bids_dir=config["bids_dir"],
         pybids_inputs=config["pybids_inputs"],
+        pybids_database_dir=config.get("pybids_db_dir"),
+        pybids_reset_database=config.get("pybids_db_reset"),
         derivatives=config["derivatives"],
         participant_label=config["participant_label"],
         exclude_participant_label=config["exclude_participant_label"]
--- a/snakebids/tests/test_generate_inputs.py
+++ b/snakebids/tests/test_generate_inputs.py
@@ -1,7 +1,9 @@
 from __future__ import absolute_import
 
+import filecmp
 import itertools as it
 import os
+import shutil
 from collections import defaultdict
 from pathlib import Path
 
@@ -9,7 +11,11 @@ import pytest
 from bids import BIDSLayout
 
 from snakebids.core.construct_bids import bids
-from snakebids.core.input_generation import _get_lists_from_bids, generate_inputs
+from snakebids.core.input_generation import (
+    _gen_bids_layout,
+    _get_lists_from_bids,
+    generate_inputs,
+)
 from snakebids.tests.helpers import BidsListCompare
 
 
@@ -303,3 +309,80 @@ def test_get_lists_from_bids():
             "t1": {"acq": "{acq}", "subject": "{subject}"},
             "t2": {"subject": "{subject}"},
         }
+
+
+class TestDB:
+    @pytest.fixture(autouse=True)
+    def start(self, tmpdir):
+        self.tmpdir = tmpdir.strpath
+
+        # Copy over test data
+        shutil.copytree("snakebids/tests/data/bids_t1w", f"{self.tmpdir}/data")
+        assert filecmp.dircmp("snakebids/tests/data/bids_t1w", f"{self.tmpdir}/data")
+
+        # Create config
+        self.bids_dir = f"{self.tmpdir}/data"
+        self.pybids_db = {"database_dir": "", "reset_database": False}
+
+    def test_database_dir_blank(self):
+        # Test non-saving (check db does not exist)
+        _gen_bids_layout(
+            bids_dir=self.bids_dir,
+            derivatives=False,
+            pybids_database_dir=self.pybids_db.get("database_dir"),
+            pybids_reset_database=self.pybids_db.get("reset_database"),
+        )
+        assert not os.path.exists(self.pybids_db.get("database_dir"))
+
+    def test_database_dir_relative(self):
+        # Update config
+        self.pybids_db["database_dir"] = "./.db"
+
+        # Check to make sure db exists (relative path)
+        _gen_bids_layout(
+            bids_dir=self.bids_dir,
+            derivatives=False,
+            pybids_database_dir=self.pybids_db.get("database_dir"),
+            pybids_reset_database=self.pybids_db.get("reset_database"),
+        )
+        assert not os.path.exists(f"{self.tmpdir}/data/.db/")
+
+    def test_database_dir_absolute(self):
+        # Update config
+        self.pybids_db["database_dir"] = f"{self.tmpdir}/data/.db/"
+        self.pybids_db["reset_database"] = False
+
+        # Check to make sure db exists (absolute path)
+        _gen_bids_layout(
+            bids_dir=self.bids_dir,
+            derivatives=False,
+            pybids_database_dir=self.pybids_db.get("database_dir"),
+            pybids_reset_database=self.pybids_db.get("reset_database"),
+        )
+        assert os.path.exists(f"{self.tmpdir}/data/.db/")
+
+        # Test reading of old layout when changes occur
+        os.makedirs(f"{self.tmpdir}/data/sub-003/anat")
+        shutil.copy(
+            f"{self.bids_dir}/sub-001/anat/sub-001_acq-mprage_T1w.nii.gz",
+            f"{self.bids_dir}/sub-003/anat/sub-003_acq-mprage_T1w.nii.gz",
+        )
+        # Check to make sure new subject not cached in layout
+        layout = _gen_bids_layout(
+            bids_dir=self.bids_dir,
+            derivatives=False,
+            pybids_database_dir=self.pybids_db.get("database_dir"),
+            pybids_reset_database=self.pybids_db.get("reset_database"),
+        )
+        assert not layout.get(subject="003")
+
+        # Test updating of layout
+        self.pybids_db["reset_database"] = True
+        # Check to see if new subject in updated layout
+        layout = _gen_bids_layout(
+            bids_dir=self.bids_dir,
+            derivatives=False,
+            pybids_database_dir=self.pybids_db.get("database_dir"),
+            pybids_reset_database=self.pybids_db.get("reset_database"),
+        )
+        assert layout.get(subject="003")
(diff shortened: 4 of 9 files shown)


snakebids/project_template/{{cookiecutter.app_name}}/config/snakebids.yml
```diff
@@ -33,11 +33,16 @@ pybids_inputs:
       - acquisition
       - task
       - run
-   
+
+#this configures the options to save the BIDSLayout
+# by default, database is not reindexed
+pybids_db:
+  database_dir: '/path/to/db_dir' # Leave blank if you do not wish to use this
+  write_database: False
```
pvandyken, line 41: I find this somewhat confusing: on first glance it looks as if `write_database == False` precludes any database from being written, even if `database_dir` is set. This is not the case: `write_database == False` only prevents a reset of the database. If not database exists, one will be created if a valid path is given.

I would resolve this one of two ways:
1. Keep existing behaviour, but rename `write_database` to `reset_database`. Comment out all three of these lines by default in order to make database use explicitly opt-in.
2. Modify the behaviour so that `write_database == False` actually does prevent any database from being written.

snakebids/tests/test_generate_inputs.py
```diff
@@ -197,3 +207,51 @@ def test_get_lists_from_bids():
             "t1": {"acq": "{acq}", "subject": "{subject}"},
             "t2": {"subject": "{subject}"},
         }
+
+
+def test_db():
+    # create config
+    real_bids_dir = "snakebids/tests/data/bids_t1w"
```
pvandyken, line 214: This test will fail if it gets interrupted and run again, as `{real_bids_dir}/sub-003/anat` won't be deleted.

I'd use the pytest fixture `tmpdir`. Start by using `shutil.copytree` to copy the entire `{real_bids_dir}` into `tmpdir`. Then proceed with the rest of the test as written, using the newly created temporary dataset. That way, the original, source-controlled test folders are never touched. Pytest will take care of cleanup (by default it saves files from the last two tests), so `shutil.rmtree` would no longer necessary.

snakebids/tests/test_generate_inputs.py
```diff
@@ -197,3 +207,51 @@ def test_get_lists_from_bids():
...
+    pybids_db = {"database_dir": "", "write_database": False}
+
+    # Test non-saving first
+    _ = _gen_bids_layout(bids_dir=real_bids_dir, derivatives=False, pybids_db=pybids_db)
+
+    assert not pybids_db.get("database_dir")
+
+    # Test saving of new layout (update config)
+    pybids_db["database_dir"] = "snakebids/tests/data/.db/"
```
pvandyken, line 223: The db should be saved to a temporary directory `tmpdir/".db"`. This is to prevent a failing test if a previous run is interrupted.

snakebids/project_template/{{cookiecutter.app_name}}/workflow/Snakefile
```diff
@@ -10,6 +10,7 @@ config.update(
     snakebids.generate_inputs(
         bids_dir=config["bids_dir"],
         pybids_inputs=config["pybids_inputs"],
+        pybids_db=config["pybids_db"],
```
pvandyken, line 13: The `pybids_db` part of config should not, in my opinion, be strictly necessary. If that is so, this line should be `config.get("pybids_db")`

pvandyken (changes requested):

snakebids/project_template/{{cookiecutter.app_name}}/config/snakebids.yml
```diff
@@ -33,11 +33,16 @@ pybids_inputs:
       - acquisition
       - task
       - run
-   
+
+#this configures the options to save the BIDSLayout
+# by default, database is not reindexed
+pybids_db:
+  database_dir: '/path/to/db_dir' # Leave blank if you do not wish to use this
```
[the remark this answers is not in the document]
kaitj: This was just following in the input / output paths in the template. We can set it to where we would like, but we should decide on where that would be. Also I can't recall, but if the default behavior we want is to not make use of this then it should actually be set to `''`

snakebids/core/input_generation.py
```diff
@@ -15,6 +15,7 @@
 def generate_inputs(
     bids_dir,
     pybids_inputs,
+    pybids_db=None,
```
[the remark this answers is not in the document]
kaitj: Open to splitting it - thought here was to db parameters together similar to how the bids inputs are together.

snakebids/project_template/{{cookiecutter.app_name}}/config/snakebids.yml
kaitj: Will change to `reset_database` - end of the day the writing is handled by `pybids`

snakebids/tests/test_generate_inputs.py
kaitj: I'm open to it, though when I tried to use the pytest fixture, the folders would be created at the start. I guess the way around it is to create a temp directory, copy the existing files over, and the make the new subject. Will look into it later this week.

snakebids/tests/test_generate_inputs.py
```diff
@@ -18,11 +23,13 @@ def test_t1w():
             "wildcards": ["acquisition", "subject", "session", "run"],
         }
     }
+    pybids_db = {"database_dir": "", "write_database": False}
 
     # Can't define particpant_label and exclude_participant_label
     with pytest.raises(ValueError) as v_error:
         config = generate_inputs(
             pybids_inputs=pybids_inputs,
+            pybids_db=pybids_db,
```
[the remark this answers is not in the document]
kaitj: Ended up just reverting this back to what it was originally, and `pybids_db` is only called in the `test_db()`

snakebids/tests/test_generate_inputs.py
```diff
@@ -197,3 +207,51 @@ def test_get_lists_from_bids():
...
         }
+
+
+def test_db():
+    # create config
+    real_bids_dir = "snakebids/tests/data/bids_t1w"
+    pybids_db = {"database_dir": "", "write_database": False}
+
+    # Test non-saving first
+    _ = _gen_bids_layout(bids_dir=real_bids_dir, derivatives=False, pybids_db=pybids_db)
```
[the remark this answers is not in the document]
kaitj: Fixed in latest commits

snakebids/project_template/{{cookiecutter.app_name}}/config/snakebids.yml
kaitj: This is now changed to `reset_database`.

snakebids/tests/test_generate_inputs.py
kaitj: Took a bit of finagling, but got this to working with `pytest.fixture`. It should be creating testing and updating in a `tmpdir` now.

snakebids/project_template/{{cookiecutter.app_name}}/workflow/Snakefile
kaitj: Updated

snakebids/core/input_generation.py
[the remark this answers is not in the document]
pvandyken: I'm fine with grouping them in the config, but I don't think they should be glued together in the function signature. The difference between the db params and bids inputs is that bids inputs can have any arbitrary keys which, in turn, contain the data. The db param structure is fixed and static, thus well suited to have explicit params in `generate_inputs`

snakebids/project_template/{{cookiecutter.app_name}}/config/snakebids.yml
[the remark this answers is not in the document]
pvandyken: Yeah, the actual file name doesn't matter too much.

I agree the default behaviour should be to have the db disabled, but I also think it would be useful to have `database_dir` prepopulated with a good path choice to set an informal standard for where the database should live. I think a good compromise here would be to comment out all three lines here, a practice commonly used in config files to showcase relevant settings. An explanatory comment can tell users to uncomment the pybids_db block if they wish to use the database

pvandyken (changes requested):

snakebids/tests/test_generate_inputs.py
```diff
@@ -197,3 +202,40 @@ def test_get_lists_from_bids():
...
+    # Test saving of new layout (update config)
+    pybids_db["database_dir"] = f"{tmpdir}/.db"
+    # Check to make sure db exists
+    _gen_bids_layout(bids_dir=bids_dir, derivatives=False, pybids_db=pybids_db)
+    assert os.path.exists(pybids_db.get("database_dir"))
+
+    # Test reading of old layout when changes occur
+    os.makedirs(f"{tmpdir}/data/sub-003/anat")
+    # tmpdir.mkdir("data/sub-003")
+    # tmpdir.mkdir("data/sub-003/anat")
```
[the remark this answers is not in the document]
kaitj: Oh, forgot to remove them, was testing out the difference between them - `tmpdir` only lets you a subdirectory a single level down.

snakebids/project_template/{{cookiecutter.app_name}}/config/snakebids.yml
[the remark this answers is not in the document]
kaitj: Yes, I like this - I've commented it out locally but will wait to push changes until we have decided on an example database path to include

kaitj:
Things to add / change:
* For relative pathing, relative to `bids_dir`
* sqlite file sould be added to a `.bidsignore`(user handled) for validation purposes -> print out warning message
* separate out `pybids_db` in the function

kaitj:
Most of the things above are now fixed. 

I didn't end up including a warning message for the relative path case - need to think about this one a bit more as it would depend on where that db is stored (e.g. in the input directory itself vs a derivatives folder down the line).

Also ended up just adding a comment about the relative pathing in the boilerplate for now. Functionally, this should be working. I haven't been able to test it on a real use case, and only been testing on the test cases that have been implemented.

snakebids/tests/test_generate_inputs.py
```diff
@@ -197,3 +202,58 @@ def test_get_lists_from_bids():
...
+    pybids_db = {"database_dir": "", "reset_database": False}
+
+    # Test non-saving (check db does not exist)
+    _gen_bids_layout(
+        bids_dir=bids_dir,
+        derivatives=False,
+        pybids_database_dir=pybids_db.get("database_dir"),
+        pybids_reset_database=pybids_db.get("reset_database"),
+    )
+    assert not pybids_db.get("database_dir")
```
tkkuehn, line 222: I'm not sure this assertion achieves anything (even if the DB were made on the filesystem, I don't think this would catch it). Not sure though! An alternative might be to check that everything in `{tmpdir}/data` was in the original test data directory.

tkkuehn (changes requested):
A couple of nitpicks but this overall looks good to me!

As far as the (default) location, having thought about it a bit more, I think it makes sense to put it somewhere in the derivatives/output folder, just because it would be good to avoid making any changes to the input directory generally.

snakebids/project_template/{{cookiecutter.app_name}}/config/snakebids.yml
```diff
@@ -33,41 +33,47 @@ pybids_inputs:
       - acquisition
       - task
       - run
-   
+
+#this configures the options to save the BIDSLayout
```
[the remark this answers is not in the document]
kaitj: Fair, this may be something worth doing in another PR - just to go through an fix any spacing issues since it seems to be mixed in a number of files. I tried to keep it consistent as best as I could, but I agree, I prefer the space after the `#` too.

snakebids/tests/test_generate_inputs.py
kaitj: Hmm, I know it will throw an error if the assertion was `assert pybids_db.get('database_dir')`, though I do agree thinking about this a bit more that this may not be the best check. I like the idea of checking against the original data directory.

snakebids/core/input_generation.py
```diff
@@ -233,6 +229,40 @@ def generate_inputs(
     return inputs_config_dict
 
 
+def _gen_bids_layout(bids_dir, derivatives, pybids_database_dir, pybids_reset_database):
+    """Create (or reindex) the BIDSLayout if one doesn't exist,
+    which is only saved if a database directory path is provided
+    """
+    if os.path.exists(bids_dir):
```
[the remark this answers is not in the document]
kaitj: > A couple of nitpicks but this overall looks good to me!
> 
> As far as the (default) location, having thought about it a bit more, I think it makes sense to put it somewhere in the derivatives/output folder, just because it would be good to avoid making any changes to the input directory generally.

I've been going in circles with the default location. We would either have to assume the outputs will follow the BIDS structure (e.g. `bids_dir/derivatives`) or know where the derivatives/output folder lives (e.g. in the case of some of the folders we have where `bids` and `deriv` are separate). 

At one point I did think it might be better for it to live in the derivativse/output folder.

snakebids/tests/test_generate_inputs.py
kaitj: I kept this check in and added another that compares the two directories after it is copied over.

snakebids/core/input_generation.py
```diff
@@ -233,6 +229,40 @@ def generate_inputs(
...
+        # Check for database_dir
+        # If blank, assume db not to be used
+        if pybids_database_dir == "":
+            pybids_database_dir = None
+        # Otherwise check for relative path and update
+        elif (
+            pybids_database_dir is not None
+            and not Path(pybids_database_dir).is_absolute()
+        ):
+            pybids_database_dir = os.path.join(bids_dir, pybids_database_dir)
```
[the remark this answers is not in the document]
kaitj: Fixed in c42049c

tkkuehn:
Some good discussion on the interface issue at the dev lunch today, here's an attempt to summarize the main questions we're left with:

- How do we activate the DB feature?
- How do we specify the path to the DB file?
  - Are relative paths allowed? If so, how do we resolve them (i.e. what are they relative to)?
- How do we specify if the database is to be reset?

The answer to each of the questions basically boils down to:

1. Via the config file.
2. Via the call to `generate_inputs` in the Snakefile.
3. Via the command line.

With the current implementation, everything is handled by the config file, and relative paths are resolved relative to the input directory. My preference would be to instead resolve relative paths relative to the output directory, but then `generate_inputs` would need another argument with the output directory, which might end up redundant.

## Potential Solutions

### Config file only, paths relative to input directory

```
# pybids_db:
#   database_dir: '/path/to/db_dir' # Leave blank if you do not wish to use this
#   reset_database: False # Change this to true to update the database
```
```
config.update(
    snakebids.generate_inputs(
        bids_dir=config["bids_dir"],
        pybids_inputs=config["pybids_inputs"],
        pybids_db=config.get("pybids_db"),
        pybids_reset_database=False,
        derivatives=config["derivatives"],
        participant_label=config["participant_label"],
        exclude_participant_label=config["exclude_participant_label"]
    )
)
```
Pros: Already implemented, conceptually simple, minimal changes to `generate_inputs` interface, backwards compatible.
Cons: Might prefer not to default to putting pybids db in the input directory

Variant: Absolute paths only (less flexible but simpler implementation and no questionable default).

### Everything in generate_inputs call

```
config.update(
    snakebids.generate_inputs(
        bids_dir=config["bids_dir"],
        pybids_inputs=config["pybids_inputs"],
        derivatives=config["derivatives"],
        participant_label=config["participant_label"],
        exclude_participant_label=config["exclude_participant_label"]
        pybids_db=config["output_dir"],
    )
)
```
Pros: Easily customizable with full use of the config dict
Cons: Need to go in and edit the workflow any time you want to change anything with the DB.

To be continued on Monday...

tkkuehn:
## Potential Solutions Continued

### CLI Only

`run.py {bids_dir} {out_dir} {analysis_level} --pybids_db {db_path} --reset_pybids_db`

Could implement this in `cli.py` as part of the standard parser or add them to the standard `parse_args` mapping in the template config.

Pros: No path ambiguity (assuming `db_path` gets resolved relative to the working directory where `run.py` is called), easy to change DB use from run to run.
Cons: Adds a lot of length to CLI invocations every run if a user always wants to use the pybids DB.

### A combination of the above

For example, could allow the pybids DB location to be defined in the config file (maybe only as an absolute path), but have a CLI argument that overrides it if present. All the pros and cons of whatever solutions are present, plus:

Pros: Flexibility for the user to use whatever works best for their use case.
Cons: Need to make clear how we handle the case where the DB location is defined differently in different places.

tkkuehn:
Idea from today's meeting: Combine config (absolute paths only) and CLI (resolve path before adding it to the config dict), with CLI taking precedence if a location is defined on both.

kaitj:
I think this is mostly ready for review (thanks @tkkuehn for the help with the CLI).

Only thing missing is testing to see if config is updated from the CLI. Still trying to figure out the best way to do this (probably making use of the mock object).

snakebids/tests/test_generate_inputs.py
```diff
@@ -239,3 +245,70 @@ def test_get_lists_from_bids():
             "t1": {"acq": "{acq}", "subject": "{subject}"},
             "t2": {"subject": "{subject}"},
         }
+
+
+def test_db(tmpdir):
```
pvandyken, line 250: Could this be broken down into a class of sub-tests? e.g. `test_no_database_when_database_dir_blank`, `test_no_database_when_database_dir_relative`, `test_database_caches_database_dir_absolute`, or something like that?

pvandyken:
`test_update_config` is intended to test the function called `update_config()` in `app.py`. The best current model for testing such changes is the existing tests in `test_app`.

snakebids/tests/test_generate_inputs.py
kaitj: See [cef943a](https://github.com/akhanf/snakebids/pull/133/commits/cef943a2617519fd0e5eddc02a47547c7475803e)

pvandyken:
It does check what args were passed to the `write_config_file` function:
```py
io_mocks["write_config"].assert_called_once_with(
    config_file=Path("/tmp/output/mock/config.yaml"),
    data=expected_config,
    force_overwrite=True,
)
```

Admittedly, `write_config_files` has no test of its own...

pvandyken:
Yeah, that's right. More precisely, its `write_config_file` that needs a test

kaitj:
> Yeah, that's right. More precisely, its write_config_file that needs a test

Yeah, I mean it probably is worth adding in a test to make sure the args are passed, but I was thinking more of the `write_config_file` test, and that should probably be a separate PR that considers more than just the DB update.

kaitj:
Think I figured it out - added the two new values `pybids_db_reset` and `pybids_db_dir` for testing to at least make sure it is properly passed along to the `write_config_file` in both bidsapp and workflow mode.

kaitj:
Brought up during the dev lunch today:

Uncomment out the config values, setting `pybids_db_dir` to `None` and adding in a separate flag/value to opt in (with a default suggested path).

tkkuehn:
@kaitj This needs a merge from main, then I'll take a last look and merge.

tkkuehn (approved):
Looks good to me! Will merge now.

snakebids/project_template/{{cookiecutter.app_name}}/workflow/Snakefile
```diff
@@ -10,6 +10,8 @@ config.update(
     snakebids.generate_inputs(
         bids_dir=config["bids_dir"],
         pybids_inputs=config["pybids_inputs"],
+        pybids_database_dir=config.get("pybids_db_dir"),
+        pybids_reset_database=config["pybids_db_reset"],
```
pvandyken, line 14: This needs to be `config.get` or it will fail on the default, commented out state

snakebids/tests/test_cli.py
```diff
@@ -116,3 +116,19 @@ def test_dash_syntax_in_config_cli_args(parser: ArgumentParser, mocker: MockerFi
     args = parse_snakebids_args(parser)
     assert args.args_dict["participant_label"][0] == "12345"
     assert args.args_dict["arg_using_dash_syntax"][0] == "7890"
+
+
+def test_update_pybids_args(parser: ArgumentParser, mocker: MockerFixture):
```
pvandyken, line 121: I think this test isn't doing anything? I'd remove it on the PR, throw it on a different branch if for future reference

snakebids/project_template/{{cookiecutter.app_name}}/workflow/Snakefile
tkkuehn: Ah in workflow mode this is true, `config["pybids_db_reset"]` gets overwritten from the CLI [here](https://github.com/akhanf/snakebids/pull/133/files#diff-7474454e76eaaba6cbcaa4ebc844258048892849c14d25afa49f0f3644422d2bR193).

snakebids/tests/test_generate_inputs.py
```diff
@@ -303,3 +309,80 @@ def test_get_lists_from_bids():
...
+
+    def test_database_dir_blank(self):
+        # Test non-saving (check db does not exist)
+        _gen_bids_layout(
+            bids_dir=self.bids_dir,
+            derivatives=False,
+            pybids_database_dir=self.pybids_db.get("database_dir"),
+            pybids_reset_database=self.pybids_db.get("reset_database"),
+        )
+        assert not self.pybids_db.get("database_dir")
```
pvandyken, line 335: Are there any actual checks on the file system here? It looks like only a trivial assertion on the `pybids-db` dict is being made

pvandyken:
Also, could you squash to 1 or two commits before merging?

snakebids/tests/test_generate_inputs.py
kaitj: Ahh, you're right - at one point it was checking the file system but not in this case. Will update this assertion.

snakebids/tests/test_cli.py
kaitj: Removed from the current PR

snakebids/tests/test_generate_inputs.py
kaitj: This should be fixed by checking the file path now

snakebids/project_template/{{cookiecutter.app_name}}/workflow/Snakefile
kaitj: Updated this to `config.get("pybids_db_reset")`

kaitj:
> Also, could you squash to 1 or two commits before merging?

@pvandyken Is there an easy way to do this (a lot of commits happened since the beginning)? Actually just pushed what I thought was squashed history that ended up messing up the git history and had to revert to f9068a5.

pvandyken:
I usually just rebase... Something like `git rebase -i HEAD~40` in this case. You can `ctrl+v` block select in vim to change all the `picks` to `squash` at the same time. And make the first one `reword`.

tkkuehn:
> > Also, could you squash to 1 or two commits before merging?
> 
> @pvandyken Is there an easy way to do this (a lot of commits happened since the beginning)?

More info on this, for reference: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#_squashing

kaitj:
Yeah, the problem is coming from the merge from main - that history is in the logs too (probably should have rebased there)

tkkuehn:
Another option that's particularly useful if you want to just squash everything into one commit and don't mind losing the commit messages: https://git-scm.com/book/en/v2/Git-Tools-Reset-Demystified#_squashing_2

kaitj:
Just managed to squash to a single commit.

Issue was the merge from main rather than a rebase. Got that resolved by just rebasing from main instead.

pvandyken:
In the spirit of rebase tips, reflog is a vital tool for recovering from rebase errors. (https://www.atlassian.com/git/tutorials/rewriting-history/git-reflog). It's saved me a few times

kaitj:
Alright, after a lot of back and forth between commits, I think finally caught and fixed most of the issues that I introduced from rebasing / squashing. Definitely my least favourite part of git :sweat_smile:

pvandyken:
Looks like the test reverted back to a single test for all three cases, rather than 3 separate tests?

kaitj:
> Looks like the test reverted back to a single test for all three cases, rather than 3 separate tests?

Should be fixed in latest commit.

pvandyken (approved):

tkkuehn (approved):
Alright, merging!

pvandyken:
@kaitj, finally got a chance to use this feature in the wild... it's soooo much faster! Definitely worth all the trouble it took.

Issue #135, opened 2022-03-16
Custom Pybids Config


pvandyken:
Pybids supports the use of custom config files to allow the indexing of custom entities. Snakebids currently doesn't support these, making it impossible to index such entities without the use of custom paths. Basic support should be as simple as adding a `config` parameter to `generate_inputs` and passing it on to `BIDSLayout`.

See below for the precipitating discussion.

### Discussed in https://github.com/akhanf/snakebids/discussions/134

<div type='discussions-op-text'>

<sup>Originally posted by **LouFel** March 16, 2022</sup>
Hello,

First of all, I wanted to thank you for developing your snakebids tool. I am a post-graduate intern trying to build a PET/MR pipeline and it has been really helpful. 

After running: 
```
config.update(
    snakebids.generate_inputs(
        bids_dir=config["bids_dir"],
        pybids_inputs=config["pybids_inputs"],
        derivatives=config["derivatives"],
        participant_label=config["participant_label"],
        exclude_participant_label=config["exclude_participant_label"]
    )
)
```
In my dataset, the PET images have a "trc" label (to specify the tracer used)
I noticed that the "trc" wildcard was missing in the updated config, in `input_wildcards`.

As I understand the way the inputs generation works is:

-  the `generate_inputs()` function generate the snakemake inputs,  
- `_get_lists_from_bids()` called inside populate input_path, input_lists, input_zip_lists, and input_wildcards,
- then along the process `_process_layout_wildcard()` is called which calls `read_bids_tags()`

The `read_bids_tags()` function checks compatibility for pybids as BIDS entities label differs from their key tags (e.g. Tracer -> trc), and I noticed that it is always using the `snakebids/resources/bids_tags.json` within `_process_layout_wildcard()` (function call is empty, and bids_tags.json is the default file). 

Finally, inside the `bids_tags.json` file the pair (label/key) `"tracer": "trc"` is missing, so I figured my issue might come from here. I might be wrong but adding the missing pair "tracer":"trc" inside the `bids_tags.json` should fix the problem. 

Thanks in advance for your help :smile: 

</div>

PR #136, opened 2022-03-16
Allow custom pybids config files

pvandyken:
Here's the implementation of #135. Includes a test. No documentation as of yet on the `pybinds_config` variable, but then none of the `pybids` related parameters have documentation currently. This will be fixed in a future release.


pvandyken:
Also, somehow a couple commits from Feb 2 got nuked. No idea how it happened, but they were still in my personal Git history and ended up being rebased to the top. So they're here now.

akhanf (approved):
haven't tried it out but the code looks good to me, as does the test!

tkkuehn (approved):
Looks good to me!

kaitj (approved):

PR #138, opened 2022-03-17
Bids inputs

pvandyken:
At long last, I have my major update to `input_generation.py`. The most important change is the new `BidsInputs`, intended to replace the current `dict` returned by `generate_inputs()`. It's API is currently exactly the same as the dict, so to migrate, all that' necessary is to replace calls as such: `inputs['input_lists']` to `inputs.input_lists`. 

By default, `generate_inputs` will still return a dict, and will log a warning saying that the default behaviour will change in an upcoming release. Users can opt-in to the new `BidsInputs` class by setting the `use_bids_inputs` parameter on `generate_inputs()` to `True`, or they can stick with the `dict` and squelch the warning by setting the parameter to `False`.

Most of the functions have been refactored either to simplify them or increase their generality. The details are all written up in my commit messages; refer to them for an orderly account. They are organized and sorted thematically (and all the throw-away commits have been squashed), so it may be easier to review this one commit at a time.

The most important refactor is on the `_process_path_override()`, now called `_parse_custom_path()`, function, which now fully supports filtering using `filter_list()`

## Checklist


snakebids/utils/utils.py
```diff
@@ -59,6 +61,7 @@ def get_match_search_func(match_list: List[Any], match_func: Callable[[Any, Any]
         Takes as a single arg a value to be matched. It will be compared to every item
         in match list using match_func
     """
+    match_list = [*match_list]
```
tkkuehn, line 64: Could this be `list(match_list)`? That would be a bit more readable imo.

snakebids/core/input_generation.py
```diff
@@ -405,68 +405,77 @@ def _parse_custom_path(
...
 
-    else:
-        pattern = f"{tag}-([a-zA-Z0-9]+)"
-        replace = f"{tag}-{{{out_name}}}"
+    wildcard_values: Dict[str, str] = {}
+
+    for wildcard in wildcards:
+        # Iterate over wildcards, slowly updating the path as each entity is replaced
+
+        bids_tags = read_bids_tags()
```
tkkuehn, line 655: I think this could be pulled out of the loop.

tkkuehn (commented):
Looks good to me! The way you structured the commits was indeed helpful for reviewing, and this change introduces a lot of clarity into the API where the original dicts got confusing. A couple of tiny nitpicks that I'm not 100% sure would be functional and aren't really necessary.

The one thing we should consider is how much we want to encourage adoption of the BidsInput class. The warning that existing users are going to be seeing a lot makes the case that users should adopt the new class ASAP, but in that case I think we should transition the template workflow and docs (i.e. the tutorial text, you've included very nice docstrings here) to the new class before merging this.

pvandyken:
@tkkuehn, that's a fair point re the warning message on transitioning. I'm happy to comment it out for now, and we can display it when we deem appropriate

pvandyken:
I've addressed @tkkuehn's comments, and added the docs for the new features.

docs/bids_function/overview.rst
```diff
@@ -57,8 +57,8 @@ Now if you want to process all inputs of a given form regardless of how their wi
             acq="{acq}",
             task="{task}",
             run="{run}",
-            **config["subj_wildcards"]
+            **inputs."subj_wildcards"
```
tkkuehn, line 60: The quotes here should be removed.

docs/tutorial/tutorial.md
```diff
@@ -243,129 +271,163 @@ The config variables we need pre-defined are as follows::
...
-Note 2: entries in the ``wildcards`` list do not have to be in your bids dataset, but if they are, then they will be converted into wildcards (i.e. ``task-{task}``) if they are in the filenames. The names for these also correspond with pybids (e.g. ``acquisition`` maps to ``acq``). 
+The ``pybids_inputs`` dict defines what types of inputs the workflow can make use of (i.e. the top-level keys, ``bold`` in this case), and for each input, how to filter for them (i.e. the ``filters`` dict), and what BIDS entities to replace with wildcards in the snakemake workflow (i.e. the ``wildcards`` dict).
 
+```{note}
+The ``filters`` dict is passed directly to the ``get()`` function in ``pybids``, and thus is quite customizable
+```
 
-To see what is going on, we can also print the config dict after calling ``generate_inputs()``::
+%  (TODO: add link)
+%  Edit: A link would be nice, but pybids docs are abysmal, so there's not much to link to!
```
tkkuehn, line 279: Minor: I would tone this down a little in the name of being nice to our fellow open source neuroimaging software developers (even in throwaway comments in our documentation source).

docs/tutorial/tutorial.md
```diff
@@ -243,129 +271,163 @@ The config variables we need pre-defined are as follows::
...
 
-The ``pybids_inputs`` dict defines what types of inputs the workflow can make use of (i.e. the top-level keys, ``bold`` in this case), and for each input, how to filter for them (i.e. the ``filters`` dict), and what BIDS entities to replace with wildcards in the snakemake workflow (i.e. the ``wildcards`` dict). 
-
-Note 1: the ``filters`` dict is passed directly to the ``get()`` function in ``pybids``, and thus is quite customizable (TODO: add link)
-
-Note 2: entries in the ``wildcards`` list do not have to be in your bids dataset, but if they are, then they will be converted into wildcards (i.e. ``task-{task}``) if they are in the filenames. The names for these also correspond with pybids (e.g. ``acquisition`` maps to ``acq``). 
+The ``pybids_inputs`` dict defines what types of inputs the workflow can make use of (i.e. the top-level keys, ``bold`` in this case), and for each input, how to filter for them (i.e. the ``filters`` dict), and what BIDS entities to replace with wildcards in the snakemake workflow (i.e. the ``wildcards`` dict).
 
+```{note}
+The ``filters`` dict is passed directly to the ``get()`` function in ``pybids``, and thus is quite customizable
```
tkkuehn, line 275: Nitpick: Add a period to the end of this note.

docs/tutorial/tutorial.md
```diff
@@ -164,67 +178,79 @@ We will also add a new variable to point to our input bold file::
...
-            pybids_inputs=config['pybids_inputs'],
-        )
+    inputs = snakebids.generate_inputs(
+        bids_dir=config['bids_dir'],
+        pybids_inputs=config['pybids_inputs'],
+        use_bids_lists=True,
     )
 
+```{note}
+Snakebids is transitioning to a new format for {func}`generate_inputs() <snakebids.generate_inputs()>`. Currently, you need to opt-in to the new features by setting `use_bids_lists=True` in `generate_inputs()`, but it will become the default in an upcoming version. We recommend all new users opt-in to maintain long term support, so the tuturial is written using the new syntax. A tutorial for the old syntax can be found on [the v0.5.0 docs](https://snakebids.readthedocs.io/en/v0.5.0/tutorial/tutorial.html#part-ii-snakebids)
```
tkkuehn, line 251: Nitpick: Add a period on the end.
```suggestion
Snakebids is transitioning to a new format for {func}`generate_inputs() <snakebids.generate_inputs()>`. Currently, you need to opt-in to the new features by setting `use_bids_lists=True` in `generate_inputs()`, but it will become the default in an upcoming version. We recommend all new users opt-in to maintain long term support, so the tuturial is written using the new syntax. A tutorial for the old syntax can be found on [the v0.5.0 docs](https://snakebids.readthedocs.io/en/v0.5.0/tutorial/tutorial.html#part-ii-snakebids).
```

tkkuehn (changes requested):
A couple of minor comments on the new documentation, then this is good to go (plus a merge from main).

docs/tutorial/tutorial.md
pvandyken: Agreed, thank you for pointing this out

kaitj (commented):
Most of the things I saw when I first glanced through it was fixed. Wasn't sure about a couple of code/text in the steps, but looks good otherwise.

tkkuehn:
Semi-unrelated to this PR: It doesn't seem like the tests are being run on every commit anymore, and they probably should be.

pvandyken:
What do you mean by "every" commit? If you push multiple commits at once, it only runs tests on the last commit

tkkuehn:
As far as I can tell, the only action run for the most recent few commits was the readthedocs build, that's all I really mean.  [658b0d5](https://github.com/akhanf/snakebids/pull/138/commits/658b0d56d77883b91372203398ea3019e0a1fdf9) appears to be the last commit that had the tests run but I'm not sure what the issue is.

pvandyken:
Ah, I see the problem. But not the reason... very strange

pvandyken:
Could it possibly relate to the fact that I changed only docs, no code?

kaitj:
It looks like it is running the tests over on the forked repo. If you look at the most recent actions in @pvandyken's fork, the tests are being run there. It might have been one of the settings that was previously changed, but not sure why it isn't also being run in the main repo.

tkkuehn (approved):

PR #139, opened 2022-03-17
Update bids_tags.json

akhanf:
## Proposed changes

Manually updated bids_tags.json against entity table in latest bids specification

 As per #134 allows use of tracer instead of trc as wildcard name. Tracer isn't in pybids yet but when it is  the name is likely to be tracer instead or trc (according to existing convention to write out abbreviations in full).

## Checklist

- [x] updated json
- [ ] could also have the bids() function use bids tags to get the ordered dictionary it needs, instead up updating that hard-coded list?


kaitj (approved):
Looks good to me.

It also looks like there might be some new PRs in the works on the `pybids` side to keep up to date with the latest tags that we should keep an eye on.

pvandyken (approved):

tkkuehn (approved):

PR #140, opened 2022-03-17
Zenodo badge

kaitj:
## Proposed changes

Inspired by the [`pybids`](https://github.com/bids-standard/pybids/releases/tag/0.14.0) releases, this adds a badge for the latest associated zenodo object to both the release-drafter and readme. 

One thing I am unsure of with this PR is with future releases. Zenodo issues a new doi for each new release that is linked to the original ("concept") doi. Currently the badge is always grabbing the "latest" version, so I'm not sure if on a new release, any existing badges get updated. There might be a way to search for the matching version, but I haven't found an easy way for that yet.

## Checklist

Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you are unsure about any of the choices, don't hesitate to ask!

- [x] Changes have been tested to ensure that fix is effective or that a feature works.
- [x] Changes pass the unit tests
- [x] I have included necessary documentation or comments (as necessary)
- [x] Any dependent changes have been merged and published

## Notes
All PRs will undergo the unit testing before being reviewed. You may be requested to explain or make additional changes before the PR is accepted.

PR #143, opened 2022-03-17
Integrate MyST parser into docs

pvandyken:
Here's my first pass at integrating MyST into the docs. I changed the README, as well as a few doc files, into Markdown. I don't think there's any need to change all the files, but this will hopefully make maintenance of the docs easier in the future.


tkkuehn (approved):
Looks good to me and confirmed that I can generate the docs locally. Thanks for this, glad not to need to keep looking up rst syntax!

kaitj:
Looks good to me as well! If only this came a few hours earlier haha :sweat_smile: - we can probably merge this in before the zenodo badge (#140) so that I can update the markdown readme there instead of the rst version.

Unrelated, @pvandyken, did you self-assign during the creation of the PR? Just curious why the auto-assigner didn't assign reviewers.

PR #145, opened 2022-03-28
Dependency updates

pvandyken:
- New version (0.15.0) of pybids
- New major version of black (v22)
- New major version of PyYAML (v6) *necessary* for latest snakemake
  versions
- New major version of pytest (v7)

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -28,7 +28,7 @@ python-versions = "!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,>=2.7"
 
 [[package]]
 name = "astroid"
-version = "2.9.3"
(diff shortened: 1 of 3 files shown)


akhanf (approved):
Good idea, was wanting to try latest snakemake in hippunfold too

kaitj:
This goes back to when we were testing the time it took to run a `poetry update`, should we be using tighter versioning (e.g. `~` instead of `^`?) Otherwise looks good to me.

tkkuehn (approved):

kaitj (approved):

kaitj:
Other then adding a label, looks good to be merged!

PR #147, opened 2022-03-28
Allow absent filters or wildcards in config entry

pvandyken:
## Proposed changes
Specifically allows both wildcards and filters to be absent from the pybids config dict.
- If no filters present, all files in the layout will be parsed.
- If no wildcards present, no entries will be included in any of the
  bids-inputs items. Generally, this will lead to the log that multiple
  files match the pybids entry, and that the first match will be picked
  (unless the user has sufficiently narrow filters)

## Checklist

## Notes


tkkuehn (approved):
Looks good to me! `BidsListCompare` makes the tests a lot cleaner, too.

kaitj (approved):
This looks good to me! 

Just a thought, have we done any testing of performance or have tests for performance?

pvandyken:
> This looks good to me! 
> 
> Just a thought, have we done any testing of performance or have tests for performance?

You mean benchmarking? I think #138 brings in a dependency to help with that. Do you think this specific PR needs a benchmark?

kaitj:
> You mean benchmarking? 

Yep (word was escaping me yesterday) - I don't think this PR necessarily needs one. Just more generally and thought had crossed my mind when I was reviewing this code. We can move this into a separate discussion thread if neccesary.

PR #149, opened 2022-03-28
Change review assignment action (no automated follow-up review request)

### LINKED ISSUE #148: Reviewers Being Reassigned after force push
## The problem

@kaitj, I don't think this is a huge issue, really just a quality-of-life thing. Looks like right now when I make a force push on to an existing PR (especially after rebasing the branch onto new changes in `main`), GH refreshes all review requests (see the history of #138 for example). Is it possible and easy to set it so that only the initial pull request triggers the review request?

kaitj:
## Proposed changes

Resolves #148. Changes such that the reviewer assignment will trigger only if there was no assignee to the pull request. If a PR needs to be reviewed again (due to changes), users will have to manually request.

## Checklist

## Notes

### DIFF
--- a/.github/workflows/assign.yml
+++ b/.github/workflows/assign.yml
@@ -6,6 +6,7 @@ on:
 jobs: 
   assign:
     runs-on: ubuntu-latest
+    if: github.event.pull_request.assignee == null
 
     steps:
       - name: Assign reviewer


pvandyken (approved):
Thanks, I personally prefer it this way

kaitj:
It's a good call - must have missed the conditional when I moved it to it's own .yml file

.github/workflows/assign.yml
```diff
@@ -6,9 +6,10 @@ on:
 jobs: 
   assign:
     runs-on: ubuntu-latest
+    if: github.event.pull_request.assignee == null
```
tkkuehn, line 9: Do we want `github.event.pull_request.requested_reviewers` here? Assignee refers to the person who's working on the PR, I think.

tkkuehn: Ah okay I forgot that this action also sets the assignee. No need to change it then!

PR #150, opened 2022-04-01
Remove explicit workflow mode

pvandyken:
## Proposed changes

I think this one is ready to go, I just need a glance at the documentation. It's mostly just deleting things, which makes predicting the consequences a bit difficult, so I'd appreciate a good look-over and possibly some beta-testing on existing workflows, if possible.

There is now no explicit, opt-in workflow mode. Instead, behaviour is entirely driven by choice of output folder. "Workflow mode" will still trigger when the output dir is the same as the snakebids root, or in the results subfolder. This simply implies that the working directory will remain in the snakebids root, meaning the config file and .snakemake folder will be saved there. All other paths will trigger the old "bidsapp" mode.

I also added a cli flag to force outputs into new directories with contents, aka any directory without a .snakebids folder. This should hopefully prevent accidental writing into unintended directories without making it impossible, while still allowing seamless re-running of the same workflow. Note that these checks will not be triggered at all in the "workflow mode" conditions.

Question regarding the docs: there's currently a section dedicated to "workflow mode" but it mostly talks about setting your output dir to be equal to the snakemake dir. Very little about actually using `--workflow-mode`, and that can easily be removed. So, should I just keep references to "workflow mode" with the understanding that it's not literally a mode you turn on, but just a consequence of choosing certain output paths? It is a convenient description...

## Checklist


snakebids/utils/output.py
```diff
@@ -357,12 +105,25 @@ def _get_snakebids_file(outputdir: Path):
         return None
 
     # If it does exist, is it empty?
-    if len([*outputdir.iterdir()]) == 0:
+    dir_size = len([*outputdir.iterdir()])
```
tkkuehn, line 108: While it was already like this, I would find `len(list(outputdir.iterdir()))` a bit more readable.

Also, why make this a variable? I don't think `dir_size` gets used anywhere else.

snakebids/cli.py
```diff
@@ -232,14 +236,30 @@ def add_dynamic_args(
 
 def parse_snakebids_args(parser: argparse.ArgumentParser):
     all_args = parser.parse_known_args()
+    if all_args[0].workflow_mode:
+        logger.warning(
+            "--worflow-mode is deprecated, and no longer has any effect. To set up a "
```
tkkuehn, line 241: Typo: `--worflow-mode` -> `--workflow-mode`

tkkuehn (commented):
Looks good to me, with a couple of nitpicks. I gave it a quick shot using an adaptation of the template workflow and it appeared to work as expected (but someone with more real workflows to try could test it more exhaustively.

As far as the documentation, I think it makes sense to maintain the references to workflow mode as long as we make it clear that it's activated by choosing `snakemake_dir` as the output directory.

snakebids/cli.py
```diff
@@ -232,14 +236,30 @@ def add_dynamic_args(
 
 def parse_snakebids_args(parser: argparse.ArgumentParser):
     all_args = parser.parse_known_args()
+    if all_args[0].workflow_mode:
+        logger.warning(
```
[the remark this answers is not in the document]
pvandyken: Workflow mode will currently make a copy of the app if you specified an outside output dir. The git comments are to indicate this is not directly supported any more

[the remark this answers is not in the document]
kaitj: I think this may need to be reworded to be more specific because I agree with @tkkuehn - it is not really clear what is meant by the recommendation.

snakebids/utils/output.py
pvandyken: I removed the variable, and switched the syntax to use more_itertools.ilen() to be more semantic

kaitj (approved):
This looks good to me. Didn't know about the `more_itertools` library, but looks useful!

pvandyken:
Just realized that the deprecated files will all still appear in the CLI help, which probably isn't important. There is a way to keep the arg but prevent it from appearing in help (so that a nice error message will print if someone tries to use it). Should we go that route instead? @tkkuehn @kaitj @akhanf

tkkuehn:
> There is a way to keep the arg but prevent it from appearing in help (so that a nice error message will print if someone tries to use it).

Sounds good to me.

Also, I may be wrong but I don't think we want the `.benchmark` directory (I assume from pytest-benchmark) checked in?

pvandyken:
> Also, I may be wrong but I don't think we want the `.benchmark` directory (I assume from pytest-benchmark) checked in?

Good catch, it's only `.gitignored` in the `bids-inputs` branch, so I must have let it in by accident.

akhanf (approved):
just tried it out and works great for me!

PR #151, opened 2022-04-07
Fix template workflow

tkkuehn:
## Proposed changes

Fix a couple of issues with the template workflow that kept it from running for me. Note: It would be good to have someone test this to make sure it wasn't some issue with my local setup that now breaks the workflow for everyone else. I've also added a test to make sure a dry-run of the generated template workflow completes successfully.

~~It would also be nice to update the release drafter to update the "Snakebids Version" in the `cookiecutter.json` -- I'll try to take a look at that before we merge this.~~ Done.

## Checklist

## Notes

### DIFF
--- a/snakebids/project_template/{{cookiecutter.app_name}}/workflow/Snakefile
+++ b/snakebids/project_template/{{cookiecutter.app_name}}/workflow/Snakefile
@@ -3,7 +3,7 @@
 import snakebids
 from snakebids import bids
 
-configfile: workflow.source_path('..config/snakebids.yml')
+configfile: workflow.source_path('../config/snakebids.yml')
 
 #writes inputs_config.yml and updates config dict
 config.update(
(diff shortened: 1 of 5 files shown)


snakebids/project_template/{{cookiecutter.app_name}}/workflow/Snakefile
```diff
@@ -29,17 +29,40 @@ wildcard_constraints:  **snakebids.get_wildcard_constraints(\
 
 rule all:
     input: 
-        expand( bids(root=config['root'], datatype='func',suffix='bold.nii.gz', desc='smooth{fwhm}mm', **config['input_wildcards']['bold']),
```
pvandyken, line 32: Why does this version not work? Did it ever work?

tkkuehn: Oh I think this part works if all subjects have the same wildcards (i.e. if you don't need input zip lists), but the new one would work either way.

kaitj:
> It would also be nice to update the release drafter to update the "Snakebids Version" in the cookiecutter.json -- I'll try to take a look at that before we merge this.      

Ahh, that is a good catch and should be fairly to implement via the `bump_action` (and maybe `release`) workflow. Would just need to add another action similar to the following, but replacing `pyproject.toml` with `cookiecutter.json` and updating the `find` and `replace` as appropriate.

```
- name: Update version in files
        uses: jacobtomlinson/gha-find-replace@master
        with:
          include: 'pyproject.toml'
          find: 'version = "(?:([0-9]+\.[0-9]+\.[0-9]+.+)|([0-9]+\.[0-9]+\.[0-9]+))"'
          replace: 'version = "${{ env.NEW_RELEASE }}-pre.${{ env.NEW_BUMP }}"'
```

tkkuehn:
I've made the change to include the template in the versioning GH actions and added a test to check if a dry run of the template app completes successfully, so this is fully ready for review now.

kaitj (commented):
Looks good to me code wise. Tiny nitpick, but is there anything that is stopping the merging of `bids_t1w` and `bids_bold` datasets for testing?

tkkuehn:
> Tiny nitpick, but is there anything that is stopping the merging of `bids_t1w` and `bids_bold` datasets for testing?

Not necessarily, I don't think. `test_generate_inputs` uses `bids_t1w`, but I think the filters used in either case would allow the two datasets to be merged.

akhanf (approved):
this looks fine to get things working again - in another PR we can make more extensive changes to the template workflow to include more best-practices we want workflows to have

kaitj:
> Not necessarily, I don't think. test_generate_inputs uses bids_t1w, but I think the filters used in either case would allow the two datasets to be merged.

Sounds good! This can be merged in as is and we can combine the test datasets down the line.

kaitj (approved):

tkkuehn:
Merging this per meeting today so @pvandyken can integrate it into an ongoing PR.

PR #157, opened 2022-05-06
Remove backward-compatible pybids func call

pvandyken:
As it turns out, we had already updated pybids to `v0.15` at some point, so here I just did automatic updating of all the other deps and removed the extra code needed in pybids `v0.14`


kaitj (approved):
This looks good to me! 

Interesting - `0.14.1` (release in April) was a backwards fix, but `0.15.0` was released in March. Probably had something to do with what versions we had if we updated the `poetry.lock` at some point.

PR #160, opened 2022-05-10
V0.5.1

pvandyken:
Collection of commits to release as `v0.5.1`. Made this after running `snakedwi` which is pinned at `0.5.x` and realized I needed some of these updates, especially the dependency updates for the latest snakemake.

Currently includes:
- #151 
- #145 
- #137 
- #139 
- c79e5526d4dd

This needs to be merged into a `v0.5.1` branch then turned into a tag. ~Someone with write access will need to do that. (The easiest way is to make a new branch from the `v0.5.0` tag that I can merge this into)~ (EDIT: as it turns out, I do have write access) @kaitj, I'm not sure what the easiest way would be to prepare release notes... perhaps we should just manually release this and copy/paste the relevant parts?


kaitj:
> I'm not sure what the easiest way would be to prepare release notes... perhaps we should just manually release this and copy/paste the relevant parts?

I can take a closer look tomorrow, but that might be the easiest way. I'm not sure how that might affect the release notes for `0.6.0`

tkkuehn (approved):
Seems good to me though I will note the branch you're trying to merge into is called `0.5.0`, not `0.5.1`. Was that intentional?

pvandyken:
> Seems good to me though I will note the branch you're trying to merge into is called `0.5.0`, not `0.5.1`. Was that intentional?

Yes, apologies for the confusion. My local branch was already called `v0.5.1`, and I didn't know how to add a new branch upstream without push a local branch, but I didn't want to directly push `v0.5.1` without a PR, so I called it `0.5.0`. I'll name the tag appropriately and delete the branch when finished.

pvandyken:
@kaitj, unless you suggest otherwise, I'll leave it to you to do the release so that you can integrate with release-drafter in whatever way seems best

snakebids/project_template/cookiecutter.json
```diff
@@ -5,5 +5,5 @@
   "app_full_name": "My BIDS App",
   "app_name": "{{ cookiecutter.app_full_name.lower().replace(' ', '_').replace('-', '_') }}",
   "app_description": "My template snakebids app.",
-  "snakebids_version": "0.4.3"
+  "snakebids_version": "0.5.0"
```
kaitj, line 8: Should this be `0.5.0` or `0.5.1`?

kaitj (commented):
Looks good to me, just the one question about the version in the cookiecutter.json

kaitj:
> @kaitj, unless you suggest otherwise, I'll leave it to you to do the release so that you can integrate with release-drafter in whatever way seems best

I think you are good to tag and release. After a closer look, I think a manual release for the CHANGELOG is still the best way to do it. We should keep an eye out for the 0.6.0 release (whenever that happens) to make sure the notes are correct or if we need to make any manual changes.

snakebids/project_template/cookiecutter.json
pvandyken: Ah, thanks, I didn't see this

pvandyken:
I've merged and set up a draft release. It just needs to be submitted, and the wheel needs to be uploaded to pypi. I don't know that the auto-deploy is set up to do this... @akhanf, will you be able to upload manually?

akhanf:
just noticed the version bump modified things -- I'll manually change them back and commit with [skip ci] to hopefully avoid it getting clobbered again

akhanf:
ok should be published now!

kaitj:
Just a note that this seemed to break the history for release-drafter and it can't seem to find the other PRs that were not part of 0.5.1. Will have to go in and manually update the notes when we make the next release - should be fine after that.

pvandyken:
It looks like the `filter-by-commitish` option may be useful here? One of the last in the [list of options](https://github.com/release-drafter/release-drafter#configuration-options)

kaitj:
Seems like that would have worked if we had set the `commitish` variable to the version target (e.g. `0.5.x`, `0.6.x`). However, the way it is currently set up, all of them target `master`/`main` by default. There is a write up here about how to use branch specific targets: https://github.com/release-drafter/release-drafter/issues/656#issuecomment-806184887.

May be good to have a discussion about releases to avoid this in the future?

pvandyken:
Thanks for the link. I pushed a new `v0.5.x` branch, and followed the steps in the guide to re-point the `v0.5.1` release to it. So it looks like if we update our workflows, it should work?

kaitj:
This was just released: https://github.blog/changelog/2022-06-07-you-can-now-manually-select-a-previous-release-when-generating-release-notes/

Don't know if it gets around the issue we were experiencing before, but a nice feature either way.

pvandyken:
That reminds me of this one, came out about a month ago. It looks like it solves our issues with branch protection? https://github.blog/changelog/2022-05-17-consistently-allow-github-apps-as-exceptions-to-branch-protection-rules/

jetersen:
> Seems like this could potentially work, wish there was an easier way to test the action.

@kaitj I have plans to buff up release-drafter.

Make a release drafter core and than have a probot, cli, github action version of it 😉 Gonna be a big v6 making it easy to test and potentially make it better than v5 and limited options you have based on GitHub actions event input.

PR #161, opened 2022-05-12
force fix version [skip ci]

akhanf:
quick fix to set version to 0.5.1 before release


pvandyken:
Just a little confused here... `v0.5.1` is ready for release on the (somewhat confusingly named) `v0.5.0` branch, so we don't need any push to main.

akhanf:
This was added in error, can disregard..

Issue #173, opened 2022-06-10
Cookiecutter vulnerability


tkkuehn:
## The problem

I got a dependabot alert in another repo for this -- Looks like there's a vulnerability in cookiecutter that we should get rid of by updating it: https://security.snyk.io/vuln/SNYK-PYTHON-COOKIECUTTER-2414281

Note that I don't think we use this functionality at all so it's probably not super high priority but would be good to address.

PR #174, opened 2022-06-10
Update cookiecutter for injection vulnerability

pvandyken:
See https://security.snyk.io/vuln/SNYK-PYTHON-COOKIECUTTER-2414281

Addresses #173


tkkuehn:
Build is failing because of this: https://status.python.org/incidents/lgpr13fy71bk

The code looks good but I will reserve approval until the test workflow runs through.

pvandyken:
Ah, I was wondering why nothing `pip` related was working

kaitj (approved):
Looks good to me - don't think it would break anything. We can consider making this a release on its own so alerts for repos making use of snakebids can also be resolved (e.g. #173).

pvandyken:
Good call, I'll go ahead and make the release

PR #175, opened 2022-06-15
BidsInputs Refactoring

pvandyken:
## Proposed changes

Split `BidsInputs` into two classes: a smaller class describing a single dataset component and containing the lists, zip_lists, wildcard dict, and path, and a wrapper class compiling them together that takes the same API as `BidsInputs`. The inner class I've called `BidsComponent` and the wrapper I've called `BidsDataset`. The alternative name for the wrapper was `BidsComponents` (pluralized), but I found it confusing from a documentation perspective, and I think `BidsDataset` is more descriptive.

This new structure cleans up the implementation details in a few ways:

1. The old internal `_BidsLists` named tuple is no longer necessary and has been removed.
2. `input_lists` and `input_wildcards` are now defined in terms of `input_zip_lists`. This will make modification of the dataset (e.g. filtering) much easier, as now only one entry needs to be modified.
3. `BidsComponent` is equipped with an `__eq__()` operator, making testing much easier. Most of the tests in `test_generate_inputs.py` have been modified to take advantage of this new behaviour

Finally, this opens up new ways of interacting with datasets. The previous syntax (e.g. `inputs.input_lists["t1w"]`) will still work, but because `BidsDataset` is implented as a dict, a reverse syntax is also possible (e.g. `inputs["t1w"].input_lists`). This new syntax is possibly more intuitive.

## Backward Compatibility

As `BidsInputs` was always the return of `generate_inputs()` and never intended to be directly accessed, this change should not affect any existing workflows. If anyone feels it necessary, it should be possible to have a stub `BidsInputs` whose `__new__()` returns a copy of `BidsDataset`.

## Checklist

### DIFF
--- a/snakebids/core/input_generation.py
+++ b/snakebids/core/input_generation.py
@@ -7,6 +7,7 @@ import operator as op
 import re
 from collections import UserDict, defaultdict
 from pathlib import Path
+from string import Formatter
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -27,6 +28,7 @@ from bids import BIDSLayout, BIDSLayoutIndexer
 from cached_property import cached_property
 from typing_extensions import Literal, TypedDict
 
+import snakebids.utils.sb_itertools as sb_it
 from snakebids.core.filtering import filter_list
 from snakebids.utils.snakemake_io import glob_wildcards
 from snakebids.utils.utils import read_bids_tags
@@ -74,6 +76,20 @@ class BidsComponent:
     input_name: str = attr.field(on_setattr=attr.setters.frozen)
     input_path: str = attr.field(on_setattr=attr.setters.frozen)
     input_zip_lists: Dict[str, List[str]] = attr.field(on_setattr=attr.setters.frozen)
+
+    @input_zip_lists.validator  # type: ignore
+    def _validate_input_zip_lists(self, _, value: Dict[str, List[str]]):
+        lengths = {len(val) for val in value.values()}
+        if len(lengths) > 1:
+            raise ValueError("input_zip_lists must all be of equal length")
+        _, raw_fields, *_ = sb_it.unpack(
+            zip(*Formatter().parse(self.input_path)), [[], [], []]
+        )
+        fields = filter(None, raw_fields)
+        if set(fields) != set(value):
+            raise ValueError(
+                "input_zip_lists entries must match the wildcards in input_path"
+            )
 
     # Note: we can't use cached property here because it's incompatible with slots.
     _input_lists: Optional[Dict[str, List[str]]] = attr.field(
--- a/snakebids/core/input_generation.py
+++ b/snakebids/core/input_generation.py
@@ -5,65 +5,36 @@ import json
 import logging
 import operator as op
 import re
-from collections import defaultdict
+from collections import UserDict, defaultdict
 from pathlib import Path
 from typing import (
+    TYPE_CHECKING,
+    Any,
     Dict,
     Generator,
     Iterable,
     List,
-    NamedTuple,
     Optional,
     Tuple,
     Union,
+    cast,
     overload,
 )
 
 import attr
 import more_itertools as itx
 from bids import BIDSLayout, BIDSLayoutIndexer
+from cached_property import cached_property
 from typing_extensions import Literal, TypedDict
 
 from snakebids.core.filtering import filter_list
 from snakebids.utils.snakemake_io import glob_wildcards
-from snakebids.utils.utils import get_match_search_func, read_bids_tags
+from snakebids.utils.utils import read_bids_tags
 
 _logger = logging.getLogger(__name__)
 
 
-_BidsLists = NamedTuple(
-    "BidsLists",
-    [
-        ("input_name", str),
-        ("input_path", str),
-        ("input_zip_lists", Dict[str, List[str]]),
-        ("input_lists", Dict[str, List[str]]),
-        ("input_wildcards", Dict[str, str]),
-    ],
-)
-_BidsLists.__doc__ = """Bids inputs from a single input descriptor
-
-Attributes
-----------
-input_name : str
-    The name of the input descriptor.
-input_path : str
-    Wildcard-filled path that matches the images for this modality.
-input_zip_lists : dict
-    Each key is a wildcard entity and each value is a list of the values found for that
-    entity. Each of these lists has length equal to the number of images matched for
-    this modality, so they can be zipped together to get a list of the wildcard values
-    for each image.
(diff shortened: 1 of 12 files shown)


snakebids/core/input_generation.py
```diff
@@ -75,19 +46,19 @@ class BidsInputsDict(TypedDict):
     subj_wildcards: Dict[str, str]
 
 
-@attr.frozen
-class BidsInputs:
-    """Bids input entities and their resolved values
+@attr.define
+class BidsComponent:
+    """Component of a BidsDataset mapping entities to their resolved values
```
tkkuehn, line 51: It would probably be good to note in the docstring that this is immutable.

snakebids/tests/test_generate_inputs.py
```diff
@@ -29,6 +31,188 @@
...
+    )
+
+
+def get_bids_path(zip_lists: Dict[str, List[str]]):
+    return bids(root=".", **{entity: f"{{{entity}}}" for entity in zip_lists})
+
+
+@st.composite
+def bids_input(draw: st.DrawFn):
+    alphabet = st.characters(whitelist_categories=["Ll", "Lu", "Nd"])
```
tkkuehn, line 69: I think you can use `BIDS_ALPHABET` here.

snakebids/core/input_generation.py
```diff
@@ -96,12 +67,119 @@ class BidsInputs:
         the Snakemake wildcard used for that entity.
     """
 
-    input_path: Dict[str, str]
-    input_zip_lists: Dict[str, Dict[str, List[str]]]
-    input_lists: Dict[str, Dict[str, List[str]]]
-    input_wildcards: Dict[str, Dict[str, str]]
+    input_name: str = attr.field(on_setattr=attr.setters.frozen)
+    input_path: str = attr.field(on_setattr=attr.setters.frozen)
+    input_zip_lists: Dict[str, List[str]] = attr.field(on_setattr=attr.setters.frozen)
```
tkkuehn, line 72: I think it would be worth adding a validator to make sure that all the zip lists are the same length. Also a validator to make sure the zip lists and input path have the same wildcards.

snakebids/core/input_generation.py
```diff
@@ -143,60 +223,38 @@ def subj_wildcards(self):
...
-        bids_lists : Iterable[BidsLists]
-            BidsLists to be converted
+        Included primarily for backward compatability with older versions of snakebids,
+        where generate_inputs() returned a dict rather than the `BidsDataset` class
 
         Returns
         -------
-        Tuple of Dicts
-            Tuple of all the dicts generated from the BidsLists
+        BidsDatasetDic
```
tkkuehn, line 233: Typo

tkkuehn (changes requested):
Overall looks good, just a couple of suggestions.

snakebids/core/input_generation.py
```diff
@@ -75,19 +46,19 @@ class BidsInputsDict(TypedDict):
...
 
 
-@attr.frozen
-class BidsInputs:
-    """Bids input entities and their resolved values
+@attr.define
+class BidsComponent:
+    """Component of a BidsDataset mapping entities to their resolved values
 
     Attributes
```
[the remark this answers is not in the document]
pvandyken: Hmm good catch... Actually, I think technically `input_lists` and `input_wildcards` should be documented under their respective properties (according to the sphinx docstring standards). I'm never fully sure how to document the attributes/parameters of an attrs class, because they're usually the same thing. My opinion for this is that since this class is not meant to be directly constructed, we should list them as attributes rather than parameters, but it's open to debate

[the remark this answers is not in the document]
pvandyken: Hmm good catch... Actually, I think technically `input_lists` and `input_wildcards` should be documented under their respective properties (according to the sphinx docstring standards). I'm never fully sure how to document the attributes/parameters of an attrs class, because they're usually the same thing. My opinion for this is that since this class is not meant to be directly constructed, we should list them as attributes rather than parameters, but it's open to debate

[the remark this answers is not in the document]
pvandyken: Cool, thanks for the link

snakebids/core/input_generation.py
pvandyken: Actually, if we wanted full immutability, we could use this [immutabledict](https://github.com/corenting/immutabledict). Do people think it's worth it?

pvandyken: I'm going to branch this into its own issue

pvandyken:
@tkkuehn I've addressed the comments you left. In addition, I've refactored and organized the testing code, moving all my custom hypothesis strategies to a new module.

snakebids/core/input_generation.py
```diff
@@ -75,33 +48,172 @@ class BidsInputsDict(TypedDict):
...
+    input_zip_lists
+    """
+
+    input_name: str = attr.field(on_setattr=attr.setters.frozen)
+    input_path: str = attr.field(on_setattr=attr.setters.frozen)
+    input_zip_lists: Dict[str, List[str]] = attr.field(on_setattr=attr.setters.frozen)
+
+    @input_zip_lists.validator  # type: ignore
+    def __(self, _, value: Dict[str, List[str]]):
+        lengths = [len(val) for val in value.values()]
```
tkkuehn, line 82: This would be a bit simpler as a set comprehension.

snakebids/core/input_generation.py
```diff
@@ -75,33 +48,172 @@ class BidsInputsDict(TypedDict):
...
+    input_path
+    input_zip_lists
+    """
+
+    input_name: str = attr.field(on_setattr=attr.setters.frozen)
+    input_path: str = attr.field(on_setattr=attr.setters.frozen)
+    input_zip_lists: Dict[str, List[str]] = attr.field(on_setattr=attr.setters.frozen)
+
+    @input_zip_lists.validator  # type: ignore
+    def __(self, _, value: Dict[str, List[str]]):
```
tkkuehn, line 81: My slight preference would be for this to have an informative name (still prepended with an underscore or two)

tkkuehn (approved):
A couple of nitpicks that you can address if you agree but I'm otherwise happy with this and don't need to see it again. Thanks again for your work on this, it's really nice!

pvandyken:
@akhanf, @kaitj, did you still want to take a look, or should I merge?

kaitj (approved):

kaitj:
> 



> @akhanf, @kaitj, did you still want to take a look, or should I merge?

Just took another look at it - looks good to me.

PR #183, opened 2022-07-28
Fix forced output into occupied directory

pvandyken:
The --force-output flag had not actually been linked to the SnakebidsArgs force property, preventing running a new app in an occupied the directory. The error was also missing a comma and was getting cut short.


tkkuehn (approved):
Good catch!

Issue #168, opened 2022-05-27
Support Pybids Query.{NONE,ANY}


pvandyken:
Pybids has a special QUERY enum with NONE and ANY values that can be used as filters. NONE means the queried field should not be defined in the BIDS path, while ANY means it must be defined but the value does not matter.

I'm thinking a very natural way of supporting this is to allow yml booleans as acceptable values in the filters dict. `False` would map to `NONE`, and `True` would map to `ANY`.

PR #191, opened 2022-09-09
Allow filtering of entities using True and False

pvandyken:
The main purpose of this PR is to implement #168. Setting a filter as True requires that the entity be present. Setting to false requires its absence. The changes are tested and the documentation of the configuration page has been updated, including a general refresh of the configuration page.

What's at least equally significant in this PR is that the tests for the feature use hypothesis to generate an entire bids dataset. In particular, it adds a number of strategies that can make valid zip_lists and BidsComponents that, if turned into paths, will be properly parsed by pybids. It also introduces the use of pyfakefs, a fully mocked version of the file system I employed specifically for the hypothesis tests to avoid generating hundreds-thousands of files on a real file system. There's a few fixtures and helpers for adaptation to the fakefs, including a special fakefs_tmpdir fixture. I also provide a means to disable the fake_fs by providing the `disable_fakefs` pytest mark. This is necessary as the vscode debugger (at least) doesn't integrate properly with fakefs, and so breakpoints stop working properly. In addition, the `debug` decorator can be applied to a hypothesis test to debug a specific failing example. It disables hypothesis example generation and the fakefs so that the debugger works correctly.

One caveat to note is that the new hypothesis tests are rather long (~30s per test). I haven't benchmarked it, but I imagine a good chunk of time is just spent on generating the pybids index (pybids is known for being somewhat slow), which is rather integral to the test. And there's four testing assertions, meaning four tests. Rolling the tests into one obviously saves on example generation time, but doesn't help the test time, and is anti-pattern for testing. In principle the solution is to make the test more granular, but I couldn't come up with an elegant way of doing that. Right now the test runs `generate_inputs()`, and while I could have the test directly run `get_lists_from_bids`, I expect the time savings would be marginal.

Long term it might be nice to have one sort of integration test that uses example generation to capture as many situations as possible (with possibly a long runtime), and have everything else be snappy unit tests. But this may require some restructering of the code.


pvandyken:
Ah yes, I also uncovered a few bugs/limitations while putting this together. I've noted them in comments at relevant spots in the code, but I'll go ahead and leave dedicated issues as well.

pvandyken:
One more thing: as of now, this PR will error out if an ambiguous path is found. This *could* be considered breaking: datasets or configuration may have to be modified for some datasets, even though no code changes should be necessary. I personally think this is acceptable, but one could argue for a gradual adoption of this rule.

tkkuehn (approved):
Looks good to me with one very minor comment! The tests run tolerably fast for me.

pvandyken:
@akhanf @kaitj , did either of you want to review this, or am I good to merge?

akhanf:
ah sorry I lost track of this and haven't reviewed -- but I'm ok with merging if you guys are happy with it

snakebids/tests/strategies.py
```diff
@@ -1,41 +1,63 @@
...
-    entities = draw(bids_entity_list())
+    if entities is None:
+        entities = draw(bids_entity_lists(min_size=min_size, max_size=max_size))
 
     values = {
-        key: draw(st.lists(bids_value(), min_size=1, unique=True)) for key in entities
+        entity: draw(
+            st.lists(
+                bids_value(entity.pattern if restrict_patterns else ".*"),
+                min_size=1,
```
kaitj, line 55: Should these (`min_size`, `max_size`) be using the variables rather than hard coded?

kaitj (commented):
This looks good to me aside from the one comment about the hard coding vs variable. Once that is resolved, good to be merged I think.

snakebids/tests/strategies.py
pvandyken: The `min_size` and `max_size` are already used for drawing entities. For the lists, `min_size` and `max_size` should indeed not be hardcoded, but the strategy will need to be rethought, or we'll end up with duplicate `min_size`/`max_size` arguments. I left a comment explaining as much and will defer the problem to another day

PR #195, opened 2022-10-03
Migrate poetry syntax to poetry v1.2

pvandyken:
Really simple one, just updating the `pyproject.toml` to poetry 1.2. This prevents us from using poetry 1.1, so if anyone needs to stay legacy, speak now :smiley:

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ This package allows you to build BIDS Apps using Snakemake. It offers:
 Contributing
 ============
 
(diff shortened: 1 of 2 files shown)


tkkuehn (approved):

kaitj (approved):
Lgtm - would be good if we can specify somewhere Poetry version somewhere (in the README?)

I've linked it with the issue so that it'll get closed when merged, but feel free to unlink it if it shouldn't be linked.

pvandyken:
Good point, I added a note

PR #198, opened 2022-10-21
Input aliases

pvandyken:
## Proposed changes

Here we lose the `input_*` portion of the attrs and properties from `BidsComponent` and `BidsDataset`. The original properties are kept as aliases and are documented as such. Thus, this is not a breaking change, however, the shorter names will be preferred going forward. The documentation has thus been completely updated to use the new names. 

A test ensures the aliases are maintaining proper identity.

One "extra" change I made: I noted privately with @akhanf and @kaitj that `BidsComponent.lists` just doesn't seem like a very good name, especially when you consider the equality:

```py
inputs.lists == {
  "bold": {
    "subject": ["001", "002"],
    "session": ["001"],
  }
}
```

The variable definitely doesn't describe a list as such, and is very similar to a Python primitive. We thought we should change the name to `entities`, retiring the existing attribute named `entities` (which was never released in a stable version anyway). This is consistent with the pybids method `get_entities()`, which returns a mapping of entity names to values. Obviously, the old `input_lists` remains the same. Any further feedback on this would be appreciated.

## Checklist


snakebids/core/input_generation.py
```diff
@@ -217,29 +223,57 @@ def __setitem__(self, _: Any, __: Any):
...
-        return {key: value.input_zip_lists for key, value in self.data.items()}
+        return self.zip_lists
 
     @cached_property
-    def input_lists(self):
+    def entities(self):
+        """Dict mapping :class:`BidsComponents <snakebids.BidsComponent>` names to \
+        to their :attr:`lists <snakebids.BidsComponent.entities>`
+        """
         return {key: value.input_lists for key, value in self.data.items()}
```
kaitj, line 252: `values.entities`?

snakebids/core/input_generation.py
```diff
@@ -108,67 +129,62 @@ def _validate_input_zip_lists(self, _, value: Dict[str, List[str]]):
...
         """
         if self._input_lists is None:
             self._input_lists = {
-                entity: list(set(values))
-                for entity, values in self.input_zip_lists.items()
+                entity: list(set(values)) for entity, values in self.zip_lists.items()
             }
         return self._input_lists
 
+    @property_alias(entities, "lists", "snakebids.BidsComponent.entities")
```
kaitj, line 144: "`entities`"? (there is a similar call below)

snakebids/core/input_generation.py
```diff
@@ -217,29 +223,57 @@ def __setitem__(self, _: Any, __: Any):
...
+    @property_alias(zip_lists, "zip_lists", "snakebids.BidsDataset.zip_lists")
     def input_zip_lists(self):
-        return {key: value.input_zip_lists for key, value in self.data.items()}
+        return self.zip_lists
 
     @cached_property
-    def input_lists(self):
+    def entities(self):
+        """Dict mapping :class:`BidsComponents <snakebids.BidsComponent>` names to \
+        to their :attr:`lists <snakebids.BidsComponent.entities>`
```
kaitj, line 250: `entities`

snakebids/core/input_generation.py
kaitj, line 252: `value.entities`

snakebids/core/input_generation.py
```diff
@@ -217,29 +223,57 @@ def __setitem__(self, _: Any, __: Any):
...
+        """
         return {key: value.input_lists for key, value in self.data.items()}
 
+    @property_alias(entities, "entities", "snakebids.BidsDataset.entities")
+    def input_lists(self):
```
kaitj, line 255: `entities`

snakebids/core/input_generation.py
```diff
@@ -770,23 +802,23 @@ def _parse_custom_path(
...
     #       only been performed on input_lists
     if len(wildcards[0]) == 0:
         _logger.error("No matching files for %s", input_path)
-        return input_zip_lists
+        return zip_lists
 
     # Loop through every wildcard name
     for i, wildcard in enumerate(wildcard_names):
         # Check if this wildcard needs to be filtered
         # Add the wildcard item to each output value, using filtering for input_lists
```
kaitj, line 817: entities

snakebids/core/input_generation.py
```diff
@@ -770,23 +802,23 @@ def _parse_custom_path(
         _logger.warning("No wildcards defined in %s", input_path)
 
     # Initialize output values
-    input_zip_lists: Dict[str, List[str]] = {}
+    zip_lists: Dict[str, List[str]] = {}
 
     # Log an error if no matches found
     # TODO: This will fail to detect filtering correctly as, up till now, it has
     #       only been performed on input_lists
```
kaitj, line 809: entities

snakebids/tests/test_generate_inputs.py
```diff
@@ -141,9 +156,7 @@ def test_input_lists_derives_from_zip_lists(
         path = get_bids_path(zip_lists)
 
         assert setify(
-            BidsComponent(
-                input_name="foo", input_path=path, input_zip_lists=zip_lists
-            ).input_lists
+            BidsComponent(name="foo", path=path, zip_lists=zip_lists).input_lists
```
kaitj, line 159: .entities?

kaitj (changes requested):
Think it looks good for the most part. There were instances where `input_lists` (or `lists`) were still being used rather than `entities`. I've tried to catch as many as I could (left as comments, but will let you determine which ones need to be changed), but may not have caught them all.

tkkuehn:
On the naming, I am okay with the naming as implemented in this PR but offer the following comments for thought:

- My only reservation about the name `entities` is that it seems to suggest that it includes every BIDS entity in the matched files, but in fact (I think) only contains the entities that were set as wildcards in `pybids_inputs`. That said, I think the concision is worth the trademark compared to something like `wildcard_entity_values`.
- I don't hate `zip_lists`, but `zippable_entities` or something similar would be clearer at a relatively minor length cost.

akhanf:
I agree with @pvandyken's 3 points against the name-changing fervour, and if we have a new expand() function then we don't care so much what these are named.. 

I do think the change from input_list to entities is warranted though.. The point @tkkuehn you raise about it not being all the entities could apply to anything in snakebids (since everything is specific to the particular query associated with it), so I'm not too bothered with that..

tkkuehn (approved):
The code changes seem pretty minimal and correct, and it looks to me like between you and Jason all the docs/comments will be updated appropriately. Thanks!

akhanf (approved):

pvandyken:
Okay, hopefully that fixes all the missed references. I'll merge tonight. I wonder if we should have another migration guide before releasing `0.7.0`? It's not likely the aliases will be deprecated any time soon, but maybe it would be helpful...?

kaitj (approved):

PR #200, opened 2022-11-03
Add support for python 3.11

pvandyken:
## Proposed changes
Just version changes. A few clarifications were needed on "built extensions" (pandas, numpy, scipy) to make sure poetry installs pre-built binaries (otherwise it picks old versions no longer supported for py311 but still supported for py37 and then pip tries to build them from scratch)

## Checklist

### DIFF
--- a/.github/workflows/assign.yml
+++ b/.github/workflows/assign.yml
@@ -1,15 +1,15 @@
-name: Assign reviewer 
+name: Assign reviewer
 
 on:
-  pull_request_target: 
+  pull_request_target:
 
-jobs: 
+jobs:
   assign:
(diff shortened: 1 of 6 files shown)


pvandyken:
Actually, we'll need to hold off on this: looks like Snakemake depends on a module using deprecated `async` behaviour that's been removed in py311. So python 3.11 is not actually compatible in Snakemake

pvandyken:
With Snakemake 7.18.2, we're good to go with py311. Just need an approval (@tkkuehn, @akhanf, @kaitj)

akhanf (approved):

PR #210, opened 2022-12-10
Transition versioning to dynamic-poetry-versioning

### LINKED ISSUE #207: Poetry Dynamic Versioning
Just discovered this poetry plugin for [dynamic versioning](https://github.com/mtkennerly/poetry-dynamic-versioning). Basically it eliminates the need for any hardcoded versions in the source code, instead it infers the latest version from git tags. It integrates into poetry so that when you build, the correct version gets used.

The nice thing is that it automatically handles development states. So if there's not a version tag on the current release, it will take the last version tag found and add `.devX+hash` to make a unique version for every commit.

I think this is worth integrating into the project: it would greatly simplify our github actions, make our versioning even more fine-grained and precise, and eliminate those dedicated "version bump" commits, removing the problem we had with branch protection.

@kaitj, this overlaps with a lot of the work you did on the gh actions... any thoughts?

pvandyken:
This [plugin](https://github.com/mtkennerly/poetry-dynamic-versioning) automatically calculates the version based on the git history and release tags, eliminating the need for bump version commits and greatly simplifying the workflow.

I tested the releasing on my own fork. So long as the workflow fetches the tags explicitly, it picks up and uses the version calculated by release-drafter just fine.

Still need to test this, but as far as I understand, no one should need to actually install poetry-dynamic-versioning unless they want to manually run `poetry build` and get the correct version. The versioning should otherwise work seamlessly (i.e. on local installs)

Resolves #207

## Checklist


pvandyken:
Switching to draft, as I think I can still do a better job of incorporating release-drafter into the version calculation.

pvandyken:
@kaitj, I was looking into incorporating the release-drafter calculated version into the dynamically generated version. As it is, the dynamic generator will always just bump the patch by one, even if `release-drafter` calculates a major or minor bump.

The easiest way to enforce communication between the two is to have release-drafter push a `vx.x.x.dev0` tag every time the major or minor version is bumped. This would give the dynamic versioner a new base.

This should be possible, but I think is unnecessarily complicated. The major and minor version will be correctly updated when the release is actually made, but I think for dev releases it's probably fine if they all accumulate on the next patch update. Semver includes no guarantees that a development version will remain stable.

Any thoughts?

kaitj:
I think that's fine if they aren't necessarily synced. The draft release notes are only available to devs with access to the repo (I believe) and the version there isn't a big deal as long as the correct version is synced and displayed on a release version. Was just easier in the original workfow to make use of the release-drafter calculated version.

PR #211, opened 2022-12-10
Remove redundant inputs from github actions

pvandyken:
No need to ask for author name, date, or branch, as these can be automatically determined from the context of the workflow

## Checklist


kaitj (approved):
Lgtm. Don’t think these were available when the actions were first created. Definitely useful!

PR #212, opened 2022-12-10
Update auto-assign action version

pvandyken:
The previous '@v1' was returning an error.

This is an attempt at fixing...


pvandyken:
I can't figure out why this isn't working. I've tried the specific version tag `@v1.2.4` and branch `@master`. Error is the same each time. If it doesn't resolve itself by Monday, we might have to find a new solution...

kaitj (approved):

PR #224, opened 2023-01-04
Skip Bids Indexing when `custom_paths` specified

pvandyken:
Skip Bids Indexing when all components have `custom_paths` specified. Also add a bunch of related tests, which incidentally covers some E2E cases.

## Checklist

### DIFF
--- a/snakebids/core/input_generation.py
+++ b/snakebids/core/input_generation.py
@@ -1,4 +1,3 @@
-# pylint: disable=too-many-lines
 """Utilities for converting Snakemake apps to BIDS apps."""
 
 import json
@@ -259,12 +258,16 @@ def generate_inputs(
     )
 
     # Generates a BIDSLayout
-    layout = _gen_bids_layout(
-        bids_dir=bids_dir,
-        derivatives=derivatives,
-        pybids_config=pybids_config,
-        pybids_database_dir=pybids_database_dir,
-        pybids_reset_database=pybids_reset_database,
+    layout = (
+        _gen_bids_layout(
+            bids_dir=bids_dir,
+            derivatives=derivatives,
+            pybids_config=pybids_config,
+            pybids_database_dir=pybids_database_dir,
+            pybids_reset_database=pybids_reset_database,
+        )
+        if not _all_custom_paths(pybids_inputs)
+        else None
     )
 
     filters = {"subject": subject_filter} if subject_filter else {}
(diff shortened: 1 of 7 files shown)


snakebids/types.py
```diff
@@ -0,0 +1,14 @@
+from __future__ import annotations
+
+from typing_extensions import NotRequired, TypeAlias, TypedDict
+
+
+class InputConfig(TypedDict):
```
tkkuehn, line 6: I _think_ making this `class InputConfig(TypedDict, total=False)` would mean you can drop the `NotRequired` on every field/key type.

tkkuehn (approved):
Looks great! A couple of tiny nitpicks, only one of which is really actionable, and neither of which need to block merging.

snakebids/resources/bids_tags.json
```diff
@@ -41,11 +41,11 @@
     },
     "run": {
         "tag": "run",
-        "pattern": "[0-9]+"
+        "pattern": "[0-9]{1,5}"
```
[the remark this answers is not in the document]
pvandyken: I should check this, but I *believe* this only affects internal tests. Some limit is necessary though, otherwise really long run indices cause a SQL error in pybids

snakebids/types.py
pvandyken: Did not know that, thanks!

snakebids/resources/bids_tags.json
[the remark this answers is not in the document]
tkkuehn: Gotcha. A quick scan suggests to me you're right re: this only affecting internal tests, though unrelated to this PR (and I'm pretty sure this has come up before), it would make a lot of sense to use `entity.pattern` [here](https://github.com/akhanf/snakebids/blob/main/snakebids/core/input_generation.py#L537).

snakebids/core/input_generation.py
```diff
@@ -580,12 +582,16 @@ def generate_inputs(
...
+    layout = (
+        _gen_bids_layout(
+            bids_dir=bids_dir,
+            derivatives=derivatives,
+            pybids_config=pybids_config,
+            pybids_database_dir=pybids_database_dir,
+            pybids_reset_database=pybids_reset_database,
+        )
+        if not _all_custom_paths(pybids_inputs)
+        else None
```
kaitj, line 594: Is the `else None` needed here? I would have thought if the if statement is not satisfied it just won't run the `_gen_bids_layout`, but I might be wrong.

kaitj (approved):
I don't have too much to add apart from @tkkuehn has already brought up. Only one minor question, but shouldn't block this from merging.

snakebids/core/input_generation.py
tkkuehn: The `else None` is needed because [conditional expressions](https://docs.python.org/3/reference/expressions.html#conditional-expressions) like this one require an `else` clause for the expression to evaluate when the `if` clause is falsy. (I think I did a poor job of explaining this but the bottom line is "yes" haha)

PR #225, opened 2023-01-04
Split Dataset and Component models into new file

pvandyken:
## Proposed changes

As per the title... the files were getting beyond 1000 lines, making pylint angry. The dataset file in particular will be getting larger as we build up the filtering apis


tkkuehn (approved):
Looks good to me.

pvandyken:
One thing before merging... I forgot to add the new folder back to the global `snakebids` namespace so that you can call `from snakebids import BidsDataset` etc. @tkkuehn, any objections to me doing this?

tkkuehn:
> @tkkuehn, any objections to me doing this?

No objections, sounds good.

PR #228, opened 2023-01-14
Block components with same names in dataset init

### LINKED ISSUE #181: `BidsDataset.from_iterable()` accepts multiple components with the same name
## The problem

This is a really "behind-the-scenes" bug I noticed while debugging something else. Shouldn't ever affect users working through the normal channels because a config file with multiple components of the same name would be ill-defined. But it caught me when accessing the API more directly in a test, so at some point I'll put a check in.

## Steps to reproduce

* Manually define multiple `BidsComponents` with the same `.name`.
* Pass them to `BidsDataset` via `.from_iterable`. Each successive `BidsComponent` will overwrite the preceding ones with the same name, so the `BidsDataset` will end up with a single member.

## Environment

* As of v0.6.2-pre.3

pvandyken:
BidsDataset.from_iterable was blindly converting the iterable of components passed into a dict. If any of these components had the same name, they would overlap in the dict (since the name is used as a key), and the first would be silently removed.

Fix this by checking for duplicate names.

Resolves #181

## Checklist

### DIFF
--- a/snakebids/core/input_generation.py
+++ b/snakebids/core/input_generation.py
@@ -17,7 +17,7 @@ from snakebids.core.filtering import filter_list
 from snakebids.exceptions import ConfigError
 from snakebids.types import InputsConfig
 from snakebids.utils.snakemake_io import glob_wildcards
-from snakebids.utils.utils import BidsEntity, BidsParseError
+from snakebids.utils.utils import BidsEntity, BidsParseError, surround
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


snakebids/core/input_generation.py
```diff
@@ -293,9 +293,15 @@ def generate_inputs(
             "silence this warning."
         )
         use_bids_inputs = False
+
+    try:
+        dataset = BidsDataset.from_iterable(bids_inputs)
+    except ValueError as err:
+        raise ConfigError(err.args[0]) from err
```
tkkuehn, line 300: I think I would just `raise ConfigError from err` if we don't have anything to add to the error message here. Or even something like "Multiple BidsComponents with identical names were somehow generated from the Snakebids config" to be a bit more explicit.

tkkuehn (approved):
One minor opinion-based comment that doesn't need to be addressed to merge this.

snakebids/core/input_generation.py
pvandyken: Yeah, I guess I wanted to make sure this would print something sensible to the end user, esp have something usable right at the bottom. But since this situation shouldn't really come up, the exact error is not a huge deal

kaitj (approved):

Issue #192, opened 2022-09-09
Wildcard Parsing Errors


pvandyken:
## The problem

There's two issues with snakemake wildcard parsing, but they're both related so I present them here together.

### Non-standard entities

Entities such as `extension` and `datatype` cannot be used as wildcards. Instead of capturing the relevant values, the user will get a list of blank strings:

```
BidsComponent(
  input_lists={
    "datatype": ["", "", "", ""],
    ...
  },
  ...
)
```

**Quick solution*: Build in special support for these entities

### Wildcard name overlap (e.g. `space` and `ce`)

The current regex fails to distinguish between wildcards such as `space` and `ce` (note that `ce` is the last two letters of `space`). As such, depending on the order of the wildcards, `ce` ends up matching the wrong entity and we get left with incorrect parsing.

**Quick solution**: Fix the regex to look for `_` or `/` or `\\` before the entity name.

## General solution

Both the problems above can be more generally solved by using the pybids patterns used to match the entities in the first place. These are readily accessible, are capable of matching any valid bids tag (including `datatype` etc), and are appropriately precise to prevent overlap issues. I would propose this as the matching mechanism moving forward.

pvandyken:
One more case I forgot to mention: paths with just suffixes (e.g. sub-x/func/foo.nii.gz) fail to be parsed by snakebids, even though pybids parses them correctly. Again, the issue is a faulty regex, which expects `_` to be before the suffix. Once again, we can directly correct this, or move over to `pybids` parsing patterns

PR #229, opened 2023-01-14
Use entity-specific parsers

pvandyken:
This corrects some of the mistakes made by the previous parser: mistakes caused by overlapping entity names (space and ce), inability to have "datatype" or "extension" as entities, and paths with just suffixes.

Now use the expanded bids_tags.json file to specifically match every tag. If a tag is given that isn't found in bids_tags.json, fall back to the previous `tag-[a-zA-Z0-1]` syntax.

Implementing this ended up being a bit more challenging than I thought. I was hoping I could just use the pybids entity info already available, but we need to *replace* the matched value with a wildcard, and there's no way of doing this without riskily trying to parse the pattern from pybids. Since I'm not willing to do that, we need to encode the missing information in our own `bids_tags.json` file.

I decided to use a "before", "match", "after" syntax in order to reduce data duplication (as opposed to having "pattern" with everything and "match" with just the core bit, as the core bit would appear in both "match" and "pattern"). I use "match" as the name instead of "pattern" so that it doesn't overlap with "pattern" in the pybids config file. I'm hoping that we can push this back to pybids as an update to their syntax, and this will keep it backwards-compatible.

Until then, we will unfortunately need to keep this list maintained. Pybids has previously updated their list of entities on patch releases, so we'll have to check in every once in a while to see if anythings changed.

Note that I snuck in a few type related and small refactoring changes. They're grouped together on their own commit.

Resolves https://github.com/akhanf/snakebids/issues/192

## Checklist

## Notes

### DIFF
--- a/snakebids/core/datasets.py
+++ b/snakebids/core/datasets.py
@@ -63,7 +63,7 @@ class BidsComponent:
         if fields != set(value):
             raise ValueError(
                 "zip_lists entries must match the wildcards in input_path: "
-                f"Path: {fields} != zip_lists: {set(value)}"
+                f"{self.path}: {fields} != zip_lists: {set(value)}"
             )
 
     @property
--- a/snakebids/core/input_generation.py
+++ b/snakebids/core/input_generation.py
@@ -557,9 +557,9 @@ def _parse_bids_path(path: str, entities: Iterable[str]) -> Tuple[str, Dict[str,
 
 def _get_lists_from_bids(
     bids_layout: Optional[BIDSLayout],
-    pybids_inputs,
+    pybids_inputs: InputsConfig,
     *,
-    limit_to=None,
+    limit_to: "list[str] | None" = None,
     **filters,
 ) -> Generator[BidsComponent, None, None]:
     """Grabs files using pybids and creates snakemake-friendly lists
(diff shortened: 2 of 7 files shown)


pvandyken:
As a bonus, this PR improves handling of unrecognized entities (which can happen if the user gives a custom pybids config). Previously they were actually just silently skipped, which is pretty confusing. I can see faults here being frustrating for users on the rare instances they come up (especially if it's because we aren't keeping the `bids_tags.json` file up to date), so I tried to write a really detailed error.

tkkuehn (approved):

kaitj:
Lgtm. 

Just a thought about checking against `pybids`. Need to think about this some more, but wondering if this is something we can set up an action to check against their entity file every time we have a PR and have it "fail" or some sort of warning if something has changed.

kaitj (approved):

pvandyken:
> Just a thought about checking against `pybids`. Need to think about this some more, but wondering if this is something we can set up an action to check against their entity file every time we have a PR and have it "fail" or some sort of warning if something has changed.

@kaitj, not opposed to this in principle... All depends how you do it. The easiest I guess would be to have a `pybids` version entry in our file and compare that against the latest pybids release (which should be easily queried via github api). Then just every PR we make sure as a chore that everything is up-to-date.

I definitely don't want to spend a ton of time on this though, I'd rather spend it on pushing this format back to pybids

PR #231, opened 2023-01-20
[FIX] change error message when no input found

Remi-Gau:
## Proposed changes

This is a minor change in an error message that says "no image found" even if the user does not use images as inputs.

It would now read: 

```python
_logger.warning("No input file found for %s", input_name)
```

## Checklist

## Notes


Remi-Gau:
feels pedantic but I did get confuse for second as my workflow only works on events.

so a more general error message may work better.

pvandyken (approved):
I agree, thanks!

PR #233, opened 2023-02-04
Expose BIDSLayout in BidsDataset

### LINKED ISSUE #223: how to access the layout of a dataset?
I have a bids apps that requires to get some "stats" about the dataset as a whole (on top of accessing some specific files).

Is there a way that snakebids "exposes" the layout so I can access it in script or should I just create a script that will reindex the dataset with pybids?

pvandyken:
Resolves #223

I ended up sticking with the much easier implementation of an `Optional` typed layout attribute. The key thing I missed in my post on the issue was that if all components have a custom path, `generate_input()` actually does not create a pybids layout (as of #224), and so layout could be `None` in actual workflows.

If app creators plan on using `layout` in their workflows, I would say it is their responsibility to make sure their users bids-format their data. If a user provides a custom path for every single component, I have no reason to believe they have anything nearly resembling a bids-spec directory, and so I don't want to attempt pybids parsing. In other words, I'm anticipating a proposal to allow forcing pybids indexing, and stating my early disagreement with that.

App makers focused on user experience should wrap their access of `layout` with an appropriate `try/except` block to ensure pybids effectively performed indexing.

## Checklist

## Notes


kaitj (approved):

pvandyken:
@tkkuehn, I'll add a section to the `Workflow` part of the docs explaining layout

pvandyken:
Add the relevant part to the docs. Also switched the python version used for building to py310, as the previous py37 was not correctly reading some of the type annotations.

docs/running_snakebids/overview.md
```diff
@@ -1,3 +1,5 @@
+(running overview)=
```
kaitj, line 1: I am drawing a blank at the moment with docs, but just want to make sure that this is intended?

kaitj (commented):
Looks good to me with one question with line 1 in `overview.md`

docs/running_snakebids/overview.md
pvandyken: Such is how you make a reference tag in markdown

kaitj: For reference: https://myst-parser.readthedocs.io/en/latest/syntax/syntax.html (see `section 7`).

tkkuehn (approved):

PR #234, opened 2023-02-04
Ensure bids() is called with identifying entities

### LINKED ISSUE #193: Providing only `datatype` to `bids()` give blank string
## The problem

This is admittedly an edge case, but something that came up during testing and probably worth blocking. Obviously, a bids path with *just* a datatype doesn't really make sense (pybids would never parse a path as such), but right now `bids()` just returns a blank string. I think it would be more appropriate to error. A realistic real-world situation may be that someone has:

```python
bids(
  datatype="func",
  **data.input_wildcards
)
```
where `input_wildcards` is accidentally empty (which obviously implies other major problems with their workflow, but carrying on the mistake here doesn't seem right).

pvandyken:
Calls to bids with just datatype and/or prefix are indistinguishable from calls to bids with just suffix. Pybids would parse these paths as having suffixes. Thus, we prevent this types of calls altogether

Resolves #193

## Checklist

## Notes

### DIFF
--- a/snakebids/core/construct_bids.py
+++ b/snakebids/core/construct_bids.py
@@ -130,7 +130,17 @@ def bids(
     """
     if not any([entities, suffix, subject, session]):
         raise ValueError(
-            "At least one of subject, session, suffix, or an entity must be supplied"
+            "At least one of subject, session, suffix, or an entity must be supplied.\n"
+            "\tGot only: "
+            + " and ".join(
+                filter(
+                    None,
+                    (
+                        f"datatype='{datatype}'" if datatype else None,
+                        f"prefix='{prefix}'" if prefix else None,
+                    ),
+                )
+            )
         )
 
     # replace underscores in keys (needed so that users can use reserved
--- a/snakebids/core/construct_bids.py
+++ b/snakebids/core/construct_bids.py
@@ -1,18 +1,19 @@
 """Utilities for converting Snakemake apps to BIDS apps."""
+from __future__ import annotations
 
 from collections import OrderedDict
 from pathlib import Path
-from typing import Optional, Union
+from typing import Optional
 
 
 # pylint: disable=too-many-arguments
 def bids(
-    root: Union[str, Path] = None,
-    datatype: str = None,
-    prefix: str = None,
-    suffix: str = None,
-    subject: str = None,
-    session: str = None,
+    root: str | Path | None = None,
+    datatype: str | None = None,
+    prefix: str | None = None,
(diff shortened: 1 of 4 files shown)


snakebids/core/construct_bids.py
```diff
@@ -1,18 +1,19 @@
...
 
 # pylint: disable=too-many-arguments
 def bids(
-    root: Union[str, Path] = None,
-    datatype: str = None,
-    prefix: str = None,
-    suffix: str = None,
-    subject: str = None,
-    session: str = None,
+    root: str | Path | None = None,
```
tkkuehn, line 11: Nitpick: I prefer the `Optional[str | Path]` syntax.

snakebids/utils/sb_itertools.py
```diff
@@ -30,6 +30,15 @@ def unpack(iterable: Iterable[T], default: Sequence[T]) -> Iterable[T]:
 
 # pylint: disable=invalid-name
 def drop(n: int, iterable: Iterable[T]) -> Iterable[T]:
```
kaitj, line 32: Should not prevent a merge, but it may be helpful to have this be slightly more descriptive. `drop` at a glance feels like it should be specific (e.g. `drop(7, range(10) -> [0, 1, 2, 3, 4, 5, 6, 8, 9, 10]`). Might be at the expense of a slightly longer name.

kaitj (approved):

snakebids/utils/sb_itertools.py
pvandyken: Interesting, I chose the name _à la_ `more_itertools.take`, which has the opposite function. As the function takes an `int` but can act on any iterable, it doesn't give me any impression it acts as a value filter rather than an index filter.

I'm electing right now to leave this, but feel free to push back

pvandyken:
Added another push updating the error message to be more detailed.

Also realized I made calls to bids with no args (`bids()`) illegal, but the docstring says such should return an empty string (`''`), so I fixed that

kaitj (approved):

snakebids/utils/sb_itertools.py
kaitj: Sounds good - I think I am more use to the `drop` in `pandas`. In any case, its not an issue. PR looks good to me otherwise.

pvandyken: Ah, that makes sense

tkkuehn (approved):

Issue #197, opened 2022-10-15
Pretty Print version of BidsDataset


pvandyken:
## The problem

I'm wondering if it's worth having a pretty-printed version of BidsDataset to replicate the experience from before the `attrs` based class (pretty-printed dicts). Unfortunately, the native `pprint` module in python is not extensible (there's no `__pprint__` method we can just add to the class). However, ipython does offer an extensible, powerful (supports variable-width displays and recursion prevention) version of `pprint` within their built-ins, and it can be extended using the `_repr_pretty_` method. We could recast exploratory parts of the tutorial to use ipython, which would allow pretty printing of the datasets similar to before. 

Just to note an idea I've rejected: we could also change the `__str__` or `__repr__` methods, but both of these present problems. `__repr__` is meant for terse, unambiguous, programmatic representations of an object, not pretty printing. `__str__` is too broad: if you have `print(f"the inputs variable == {inputs}")`, do you really want a large, multi-line output here?

tkkuehn:
I'm not opposed to customizing `__str__` for this (by analogy to how numpy arrays and pandas dataframes print), but there's not much of a downside to implementing `_repr_pretty_`, so that sounds good to me regardless.

tkkuehn:
We confirmed to day that numpy arrays' `__str__` method includes newlines, so we can do that too.

PR #251, opened 2023-02-18
Add deprecation warnings to `BidsDataset` properties slated for change

### LINKED ISSUE #237: Deprecate `BidsDataset` attributes slated for breaking changes
As outlined in #209, part of the `BidsDataset` API will be subject to a massive change in an upcoming release. The following attributes thus need to be deprecated at earliest opportunity to give devs time to update:

* `BidsDataset.path`
* `BidsDataset.wildcards`
* `BidsDataset.entities`
* `BidsDataset.zip_lists`

pvandyken:
Resolves #237.

This PR does not yet include any changes to the docs, as the rewrite will be a bit complicated, and I may want to incorporate the pretty printing #197 into it to ease the explanations.

I'm using the [deprecation](https://github.com/briancurtin/deprecation) library, or rather, a custom fork of it. I wanted the deprecations to be displayed prominently in the docs, and the default sphinx directive just displays it as vanilla text without any box or anything. So I modified deprecation to allow me to nest the deprecation message in a warning box. I've raised an [issue](https://github.com/briancurtin/deprecation/issues/62) there about incorporating this into the main repository.

One other note: by default, Python actually doesn't display `DeprecationWarnings`. This is to prevent end-users of an app from seeing warnings that don't pertain to them. For example, if Hippunfold upgrades `snakebids` to this PR but doesn't make the syntax changes in their workflow, there's no reason for their users to see all the resulting warning messages.

In order to the warnings they have to be enabled. Test suites do this by default. You can also call your script as `python -Wdefault`.

The problem is I don't know that snakemake users are typically using test suites on their workflows, and would probably never think to use the `-W` flag. So probably no one will ever see the warnings.

So we can just rely on the docs and changelogs to get people to migrate. There's also a separate `FutureWarning` that is always displayed, but then end users of the app will also see it.

Any ideas or thoughts on this are appreciated.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -234,6 +234,24 @@ optional = false
 python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
 
 [[package]]
+name = "deprecation"
+version = "2.1.0"
+description = "A library to handle automated deprecations"
+category = "main"
+optional = false
+python-versions = "*"
+develop = false
+
+[package.dependencies]
+packaging = "*"
+
+[package.source]
+type = "git"
+url = "https://github.com/pvandyken/deprecation.git"
+reference = "updates"
+resolved_reference = "8309c44112f349df9c38cbbfea6aed73f99ff66c"
+
+[[package]]
 name = "dill"
 version = "0.3.6"
 description = "serialize all of python"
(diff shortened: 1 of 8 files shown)


tkkuehn (approved):
Looks good to me, thanks!

pvandyken:
@tkkuehn, great, so no concern about users not seeing the warnings?

tkkuehn:
Yeah, my sense is that it's more important for end users not to get drowned in a bunch of error messages they can't address than for maintainers to see these warnings. I think updating the docs should be sufficient for anyone who's likely to eventually make the switch to Snakebids 1.0.0 (and we'll point them to the changes increasingly more clearly as we approach release of 1.0.0).

kaitj (approved):

pvandyken:
Great, just a few text changes I noticed that I'd like to fix before merging

PR #252, opened 2023-02-21
Introduce PybidsError for cryptic pybids errors

### LINKED ISSUE #159: Cryptic error messages when pybids parsing fails
The error is often something like:
```
AttributeError in line 13 of /usr/local/lib/python3.8/site-packages/funcmasker_flex/workflow/Snakefile:
'NoneType' object has no attribute 'get'
```

Would be good to catch these and try to provide a more specific error message about what happened.

tkkuehn:
## Proposed changes

This is an attempt to fix #159. I implemented @pvandyken's suggestion of raising a custom error pointing at Pybids as the root problem, but when writing the test I found it a bit hard to figure out how a Snakebids user would encounter this error in the course of normal operation (maybe @akhanf could point me at an example?).

So, the test ended up looking a bit kludgey/unrealistic, but does cover the new code.

## Checklist

## Notes

### DIFF
--- a/snakebids/core/input_generation.py
+++ b/snakebids/core/input_generation.py
@@ -14,7 +14,7 @@ from typing_extensions import Literal
 
 from snakebids.core.datasets import BidsComponent, BidsDataset, BidsDatasetDict
 from snakebids.core.filtering import filter_list
-from snakebids.exceptions import ConfigError
+from snakebids.exceptions import ConfigError, PybidsError
 from snakebids.types import InputsConfig
 from snakebids.utils.snakemake_io import glob_wildcards
 from snakebids.utils.utils import BidsEntity, BidsParseError, surround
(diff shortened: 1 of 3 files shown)


snakebids/core/input_generation.py
```diff
@@ -677,3 +685,7 @@ def get_wildcard_constraints(image_types):
         for imgtype in image_types.keys()
         for entity in image_types[imgtype].keys()
     }
+
+
+class PybidsError(Exception):
```
pvandyken, line 690: We do have `snakebids.exceptions`, maybe a better idea to put the Exception in there?

snakebids/tests/test_generate_inputs.py
```diff
@@ -856,6 +858,11 @@ def test_get_lists_from_bids():
         },
     }
 
+    broken_inputs = deepcopy(pybids_inputs)
```
pvandyken, line 861: Can this be made into a separate test?

pvandyken (changes requested):
Looks good. I actually don't understand why your test causes a failure... Because in my experience you *can* set the scope of a component to raw without error.

tkkuehn:
> I actually don't understand why your test causes a failure

I don't understand what the underlying pybids issue is, but it seems (from https://github.com/bids-standard/pybids/issues/788) that the recipe for the error is `BIDSLayout.get` with any `scope` other than `'all'`, and no `dataset_description.json` in the dataset, which are both satisfied in this test.

pvandyken:
> > I actually don't understand why your test causes a failure
> 
> I don't understand what the underlying pybids issue is, but it seems (from [bids-standard/pybids#788](https://github.com/bids-standard/pybids/issues/788)) that the recipe for the error is `BIDSLayout.get` with any `scope` other than `'all'`, and no `dataset_description.json` in the dataset, which are both satisfied in this test.

Maybe a good idea to make a note in that in the test. If `pybids` ever fixes the issue, I presume the test will fail? So we don't want to get completely confused by that.

pvandyken (approved):

kaitj (approved):

Issue #222, opened 2023-01-04
integrate bids validator by default into snake bids


Remi-Gau:
A lot of downstream errors can be prevented by early validation.

Would be useful to have an easy way for app creator to enforce validation.

Ideally it would make more sense to have validation be an opt out thing where app users have to add a `--skip-validation` flag.

For the moment this would only apply to raw datasets.

pvandyken:
I believe this is easily implementable, we'd just transparently pass on a `validate` parameter from `generate_inputs` into `BIDSLayout`. The one gotcha would be that if all `custom_paths` are provided (after #224), no `BIDSLayout` would be generated so `validate` would be ignored. But if the user cared about a valid dataset, I don't think they'd be using all `custom_paths`.

Remi-Gau:
Another option (not mutually exclusive) is to make a system call to the bids-validator (the in case the user user npm installed it.

Maybe first try the system call and if there is no bids-validator on the system then fall back to pybids validation (which I belive is just a call to the python "implementation" of the validator).

why am I suggesting a system call to the node.js validator?
it makes a lot more sense if you assume that people are going to use containerized bids apps rather than install snakebids generated apps with pip install.

akhanf:
I'd also be in favour of having validation performed as the default, and skipped as an  option --  the app developer could always override this if it was exposed in generate_inputs.

pvandyken:
@Remi-Gau, I'm not super familiar with either of the validators; what is the difference between the node version and python version? One of comprehensiveness?

Remi-Gau:
yes exactly

the python version is not a very "deep" validation. Mostly that the filenames follow bids specification

the node/dyno versions go deeper and check json / tsv file content, nifti headers...

Issue #256, opened 2023-02-22
Automatically create pybidsdb


pvandyken:
In recent versions of `snakemake`, the app will start subinstances of snakemake under many different circumstances, including cluster submission, use of the `run` directive, use of `shadow`, and probably others. Every time these sub-instances are started, `generate_inputs` will be called and, if no database exists, the dataset will be indexed. For large datasets, this is a huge problem, as indexing can take an immense amount of time (even if just a few minutes, it's still multiplied over the hundreds of times `generate_inputs` could potentially be called). Thus, for large datasets and complex workflow, a pybids database is essentially mandatory.

We could just expect users to make their own pybids database, but for downstream users, this is inconvenient. I thus propose a mechanism for automatically making a database in the case no database is provided.

This would probably conceive of this as a snakebids "extra" (see #255), so it's something app developers would need to opt into. Basically, if no database is provided, then every time `run.py` is run, it will use pybids to index the dataset and put the database into the `.snakebids` folder (which we'd need to create) in the output dir. This database will be passed on to snakemake and `generate_inputs()`.

Importantly, every time `run.py` is called, the database will be regenerated, regardless of whether it exists. This re-creates the expectation that the dataset will be re-indexed every time the app is run.

Thus, if users want a persistent database across runs, they'll still need to make their own, but within a single run, we can take advantage of the speed boosts of a database.

One concern is *how* exactly to make the db. On NAS, this can be *extremely* slow. On AllianceCan, the `/tmp` dir is mounted in memory, and I would presume in general that `/tmp` dirs are on more readily accessible storage locations. So it might be enough to always make the database on `/tmp` then copy it to `.snakebids/`.

Issue #54, opened 2021-09-13
dataset_description.json generation


pvandyken:
I would imagine the commonest use-case of this library is to generate a new derivative dataset from an existing BIDS dataset. BIDS requires a `dataset_descriptions.json` to be valid, and it would be handy if snakebids could handle this automatically. I noticed a previous commit #9 that seemed to handle this, but the code no longer seems to be present. Was there an issue in getting the implementation working correctly, or am I just missing the feature? If it is present, it doesn't seem to be documented

akhanf:
I think that feature should still be there, it is now in the cookiecutter template that is [here](https://github.com/akhanf/snakebids/blob/main/snakebids/project_template/%7B%7Bcookiecutter.app_name%7D%7D/%7B%7Bcookiecutter.app_name%7D%7D/pipeline_description.json).

But yes we should add this to documentation if it isn't there

pvandyken:
I think this is what's causing my confusion: https://snakebids.readthedocs.io/en/main/bids_app/workflow.html
>```
> import snakebids
>from snakebids import bids
>
>configfile: 'config/snakebids.yml'
>
>#writes inputs_config.yml and updates config dict
>config = snakebids.generate_inputs_config(config)
>
>#this adds constraints to the bids naming
>wildcard_constraints:  **snakebids.get_wildcard_constraints(config)
>```
In the current main branch, there is no function called generate_inputs_config(). And that seemed to be where the dataset_description code was. Has this been refactored? I can't find any similar code anymore on search.

akhanf:
Ah yes you're right, the code to copy in the .json file isn't there anymore, was left out when refactoring `generate_inputs_config()` into `generate_inputs()`.. 

I am thinking it got left out because it didn't quite fit in with what generate_inputs() does or returns (ie just returning the configuration dictionary). I am thinking it may be better add this functionality into `run_snakemake()` in `app.py` instead, since that is where the updated config is written, what do you think?

pvandyken:
Yes, that makes a lot of sense. I'll look into the old code, see if it can be put back in relatively easily, and make a PR.

PR #259, opened 2023-02-28
BIDS validation

kaitj:
## Proposed changes

An attempt at implementing bids validation (see #222). As suggested, this first attempts to perform a system call to the bids-validator prior to running `_gen_bids_layout`. If this first attempt fails, it will fall back running the validation via pybids. Validation, in both cases, are skipped if all custom paths are given. I wasn't quite sure about how to go and add a test for the system call and haven't had a chance to test this locally. There are tests for pybids.

Drawing some inspiration from fmriprep, a temporary config file is generated for the system call in (`_validate_bids_dir`) which includes files to ignore during this validation. Was also wondering if we want to limit the validation to the subject data themselves rather than a full dataset (e.g. ignore `participants.tsv, dataset_description.json`)?

_Note to self_:
- [ ] `pybids` validation
- [ ] Add to snakebids app
- [ ] Tests

## Checklist

## Notes

### DIFF
--- a/docs/bids_app/workflow.rst
+++ b/docs/bids_app/workflow.rst
@@ -14,10 +14,10 @@ To get access to these additions, the base Snakefile for a snakebids workflow sh
     inputs = snakebids.generate_inputs(
         bids_dir=config["bids_dir"],
         pybids_inputs=config["pybids_inputs"],
+        skip_bids_validation=config["skip_bids_validation"],
         derivatives=config.get("derivatives", None),
         participant_label=config.get("participant_label", None),
         exclude_participant_label=config.get("exclude_participant_label", None)
-
     )
 
     #this adds constraints to the bids naming
--- a/docs/running_snakebids/overview.md
+++ b/docs/running_snakebids/overview.md
@@ -19,6 +19,8 @@ Indexing of large datasets can be a time-consuming process. Snakebids, through `
 1. Uncomment the lines in `snakebids.yml` containing `pybids_db_dir` and `pybids_db_reset`.
 1. The variables can be updated directly in this file or through the CLI by using `-pybidsdb-dir {dir}` to specify the database path and `--reset-db` to indicate that the database should be updated. _Note: CLI arguments take precendence if both CLI and config variables are set._
 
+Input BIDS datasets are also validated via the bids-validator. By default, this feature uses the command-line (node.js) version of the [validator](https://www.npmjs.com/package/bids-validator). If this is not found to be installed on the system, the `pybids` version of validation will be performed instead. To opt-out validation, one can invoke `--skip-bids-validation`. 
+
 Workflow mode
 =============
 
--- a/snakebids/app.py
+++ b/snakebids/app.py
@@ -171,6 +171,8 @@ class SnakeBidsApp:
             self.config["pybids_db_dir"] = args.pybidsdb_dir
         self.config["pybids_db_reset"] = args.reset_db
 
+        self.config["skip_bids_validation"] = args.skip_validation
+
         update_config(self.config, args)
 
         # First, handle outputs in snakebids_root or results folder
--- a/snakebids/app.py
+++ b/snakebids/app.py
@@ -171,7 +171,7 @@ class SnakeBidsApp:
             self.config["pybids_db_dir"] = args.pybidsdb_dir
         self.config["pybids_db_reset"] = args.reset_db
 
-        self.config["skip_bids_validation"] = args.skip_validation
+        self.config["skip_bids_validation"] = args.skip_bids_validation
 
         update_config(self.config, args)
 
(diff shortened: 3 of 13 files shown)


pvandyken (changes requested):
Thanks @kaitj. I should have made this opinion known earlier, but I was thinking this type of validation might be better to do within `run.py` (*à la* #255).

There's two main reasons for this: first, it removes it from the snakebids "core", keeping things a bit more modular and hopefully easier to maintain. The second is more technical: as I mentioned in #256, `generate_inputs()` can potentially be called hundreds of times in a workflow, and running the full validator every single time would not be a great idea. So we could have tons of logic to figure out if this is the "first" run, or we could move validation to `run.py` and it would inherently only be run once.

This still leaves some implementation ambiguity: it could be a stand-alone function you import and call, or it could be part of `Snakebids` app. I'm not opinionated either way. Favouring the functional approach, it would leave space for potential future complexity (maybe you want to be able to pass a custom path to the validator executable). 

Anyway, that's my opinion, so you can let me know what you think.

pvandyken:
One more thought... this type of validation seems most relevant to containers, where the developer can guarantee `bids-validator` will be installed. For a python installation, where `bids-validator` will likely not be installed (or potentially installed in a less accessible way, like a container), a developer may want more customized ways of handling validation (skip by default, print a warning only when running the app for the first time (e.g. outputs don't exist), link to installation instructions, etc). This type of thing becomes easier when validation is more of an external module.

pvandyken:
Thought on implementation: maybe the most flexible approach would be to have a `BidsValidator` class containing the basic configuration options (executable, ignore files, config file, schema, etc), all with sensible defaults. This class gets passed to `SnakebidsApp` (maybe via a `SnakebidsApp.add_plugin()`?), which runs validation in `run_snakemake()`, providing the bids directory and other cli parameters.

Just looking through the source, it looks like we haven't actually standardized a lot of the basic CLI parameters like source_dir, output_dir, etc (they're still provided in the config). So this might require some architecture change.

kaitj:
> There's two main reasons for this: first, it removes it from the snakebids "core", keeping things a bit more modular and hopefully easier to maintain. The second is more technical: as I mentioned in https://github.com/akhanf/snakebids/issues/256, generate_inputs() can potentially be called hundreds of times in a workflow, and running the full validator every single time would not be a great idea. So we could have tons of logic to figure out if this is the "first" run, or we could move validation to run.py and it would inherently only be run once.

Just some initial thoughts, and can discuss more during a dev lunch. I think moving it to a separate module is fine. I also think it should be an opt-out feature however, so this poses a potential problem. Would also have need to have a way to differentiate between a "re-run" vs a "first-run" with additional data, which in the latter case, I would imagine you would still want to perform a validation.

>Thought on implementation: maybe the most flexible approach would be to have a BidsValidator class containing the basic configuration options (executable, ignore files, config file, schema, etc), all with sensible defaults. This class gets passed to SnakebidsApp (maybe via a SnakebidsApp.add_plugin()?), which runs validation in run_snakemake(), providing the bids directory and other cli parameters.

I like this idea. The tricky bit I had trouble with initially was figuring out an implementation that falls back on using the pybids implementation if the system call fails. At the moment, we rely on whats already implemented in pybids, but I didn't realize there was a `bids-validator` python library. I think if we can use that directly for validation, bypassing pybids (i.e. set `validate=False` always in `_gen_bids_layout`). We can still have the first pass be to the system call, with this as the fallback in this plugin.

pvandyken:
> Would also have need to have a way to differentiate between a "re-run" vs a "first-run" with additional data, which in the latter case, I would imagine you would still want to perform a validation.

True. Just to clarify, that aspect was just spit-balling. The core issue is multiple validations within a single snakemake run, which would occur if validation was done in `generate_inputs()`. Anything else is just trying to keep space for future flexibility.

> but I didn't realize there was also bids-validator python library. 

The python library, AFAIK, does no more validation than pybids does. It's probably possible to install a node library with pip, but doubtfully advisable

pvandyken:
> I also think it should be an opt-out feature however, so this poses a potential problem.

Should still be possible within my scheme, although the api for opting out might be messy.

I think ultimately we need a broader discussion about handling these types of extras (perhaps jumping off #255). There's a number of things we could do to make `snakebids` apps work better: validation, `dataset_description.json` copying #54, automatic pybidsdb #256, logging #184, and there's potentially any number of more plugin-like extras people could imagine. It would be good to have a uniform way of handling these, especially since many will require information parsed from the CLI, so we'll need the cooperation of `SnakeBidsApp`. Once we have that framework, it will be easier to decide which things should be "loaded" by default, and how that should occur.

kaitj:
> The python library, AFAIK, does no more validation than pybids does. It's probably possible to install a node library with pip, but doubtfully advisable.

Just for clarification, pybids calls the python library (I believe), so the validation should be the same. I agree about (not) trying to install the node library via pip. This is more of option to ensure that validation is still performed if it hasn't been opted out of, and if there is node library installed.

pvandyken:
Just opened #260. The goal is to use the architecture there to implement this.

kaitj:
Closing this in favour of #291.

Issue #260, opened 2023-03-03
Basic Plugin Architecture


pvandyken:
This is one of the practical manifestations of #255.

First, a basic overview of how `SnakeBidsApps` work:

```
SnakeBidsApp()                               .run_snakemake()     ??.add_plugins()
|                                            |                    |
|                                            |                    |
|-> Parse App Directory -> Create CLI parser |-> Parse arguments  |-> Run snakemake
    - locate Snakefile	   - Start with basic    - Update config
      and config             basic parser and      with the parsed
    - load config into       add items from        cli
      dict                   config          ^                     ^
                                             |                     |
			    			    Intervention Point 1      Intervention point 2

Intervention Point 1 - Available
Intervention Point 2 - Unavailable
```

The `.run_snakemake()` method gives access to intervention point 1, but point 2 is not currently accessible. There are many potential tools that need access to point 2, however, because it's only there that we get access to the parsed CLI arguments loaded into config.

Note that point 1 is still necessary for `boutiques`, which needs the constructed but unparsed CLI parser.

I propose the `SnakeBidsApp.add_plugins()` method, according to the following signature:

```py
def SnakeBidsApp.add_plugins(
	self,
	plugins: list[Callable[[SnakeBidsApp], SnakeBidsApp | None]],
) -> SnakeBidsApp:
	"""Supply list of methods to be called after cli parsing

	Each callable in ``plugins`` should take, as a single argument,
	a reference to the ``SnakeBidsApp``, and can return a new or
	modified copy of the ``SnakeBidsApp`` or ``None``. Plugins may
	perform any arbitrary side effects, including validation,
	optimization, other other enhancements to the snakebids app.

	CLI parameters may be read from ``SnakeBidsApp.config``. Plugins
	are responsible for documenting what properties they expect to find
	in the config.
	"""
```

Plugins provided via this method would be called after cli parsing and config updating, just before snakemake is invoked.

This architecture would be of immediate benefit to #259 and #256.

tkkuehn:
Spent some time looking at a basic implementation of this today -- for now I'll just note that I don't think it makes sense for plugins as defined here to return a modified `SnakeBidsApp`. To use the returned `SnakeBidsAp`p in the current framework, I think we would have to do reassign `self` within the context of `run_snakemake`, which I think might work but feels wonky. Maybe I'm not looking at this the right way though.

```
class SnakeBidsApp:
    ...
    run_snakemake(self):
        ...
        for plugin in self.plugins:
[...]

PR #261, opened 2023-03-06
Implement first pass at simple plugins

tkkuehn:
## Proposed changes

This is a very simple first crack at #260. It's entirely possible there's some potential complexity I'm missing, so I wanted to open this up for comment before adding to the Sphinx docs. Please feel free to take a look despite the fact that this is in draft mode.

## Checklist

## Notes

### DIFF
--- a/snakebids/app.py
+++ b/snakebids/app.py
@@ -1,10 +1,12 @@
 """Tools to generate a Snakemake-based BIDS app."""
+from __future__ import annotations
 
 import argparse
 import logging
 import sys
+from collections.abc import Sequence
 from pathlib import Path
-from typing import Any, Dict, List, Optional
+from typing import Any, Callable, Dict, List, Optional
 
 import attr
 import boutiques.creator as bc
@@ -111,6 +113,22 @@ class SnakeBidsApp:
         takes_self=True,
     )
     args: Optional[SnakebidsArgs] = None
+    plugins: list[Callable[[SnakeBidsApp], None]] = attr.Factory(list)
+
+    def add_plugins(self, plugins: Sequence[Callable[[SnakeBidsApp], None]]):
+        """Supply list of methods to be called after cli parsing
+
+        Each callable in ``plugins`` should take, as a single argument, a
+        reference to the ``SnakeBidsApp``, and should not return anything.
+        Plugins may perform any arbitrary side effects, including validation,
+        optimization, other other enhancements to the snakebids app.
+
+        CLI parameters may be read from ``SnakeBidsApp.config``. Plugins
+        are responsible for documenting what properties they expect to find
+        in the config.
+        """
+        # pylint: disable=no-member
+        self.plugins.extend(plugins)
 
     def run_snakemake(self):
         """Run snakemake with that config.
@@ -186,6 +204,10 @@ class SnakeBidsApp:
             cwd = args.outputdir
             new_config_file = args.outputdir / self.configfile_path
             self.config["root"] = ""
+
+        # pylint: disable=not-an-iterable
+        for plugin in self.plugins:
+            plugin(self)
 
         # Write the config file
         write_config_file(
--- a/snakebids/tests/test_app.py
+++ b/snakebids/tests/test_app.py
@@ -117,11 +117,18 @@ class TestRunSnakemake:
             reset_db=True,
         )
 
+        def plugin(my_app):
+            my_app.foo = "bar"
+
+        app.add_plugins([plugin])
+
         try:
             app.run_snakemake()
         except SystemExit as e:
             print("System exited prematurely")
             print(e)
+
+        assert app.foo == "bar"
 
         # First condition: outputdir is an arbitrary path
         if root not in ["app", "app/results"] or (root == "app" and tail):


snakebids/app.py
```diff
@@ -187,6 +205,10 @@ def run_snakemake(self):
             new_config_file = args.outputdir / self.configfile_path
             self.config["root"] = ""
 
+        # pylint: disable=not-an-iterable
+        for plugin in self.plugins:
```
pvandyken, line 209: Left my comments RE here in the linked issue

snakebids/app.py
```diff
@@ -111,6 +113,22 @@ class SnakeBidsApp:
         takes_self=True,
     )
     args: Optional[SnakebidsArgs] = None
+    plugins: list[Callable[[SnakeBidsApp], None]] = attr.Factory(list)
+
+    def add_plugins(self, plugins: Sequence[Callable[[SnakeBidsApp], None]]):
```
pvandyken, line 118: I think an `Iterable` would suffice here? `Sequence` gives "stability" (support multiple iterations) and length, and I don't think we need either of those

docs/bids_app/plugins.md
```diff
@@ -0,0 +1,5 @@
+# Plugins
+
+Plugins are a Snakebids feature that allow you to add arbitrary behaviour to your Snakebids app after CLI arguments are parsed but before Snakemake is invoked. For example, you might add BIDS validation of an input dataset to your app via a plugin, so your app is only run if the input dataset is valid.
+
+A plugin is simply a function that takes a {class}`SnakeBidsApp <snakebids.app.SnakeBidsApp>` as input and returns either a modified {class}`SnakeBidsApp <snakebids.app.SnakeBidsApp>` or `None`. To add one or more plugins to your {class}`SnakeBidsApp <snakebids.app.SnakeBidsApp>`, use the method {func}`add_plugins <snakebids.app.SnakeBidsApp.add_plugins>`, which will typically be easiest in a `run.py` or similar entrypoint to your app. Your plugin will have access to CLI parameters via `SnakeBidsApp.config`, but you're responsible for making sure that your app provides the properties a plugin expects to find in the config.
```
pvandyken, line 5: The last clause is a little confusing... I might phrase it more like the plugin has access to the config after the cli arguments have been parsed, meaning you can access cli args via their names in `bidsapp.config`.  You can also highlight that any modifications to `bidsapp.config` will be carried forward into the workflow.

As a bonus, using the (currently potential) bids validator as an example would demonstrate how the plugin can read the input directory and a skip_validation parameter out of `bidsapp.config`

snakebids/app.py
```diff
@@ -111,6 +113,32 @@ class SnakeBidsApp:
...
+
+    def add_plugins(
+        self, plugins: Iterable[Callable[[SnakeBidsApp], None | SnakeBidsApp]]
+    ):
+        """Supply list of methods to be called after cli parsing
+
+        Each callable in ``plugins`` should take, as a single argument, a
+        reference to the ``SnakeBidsApp``, and should return either:
+
+            - Nothing, in which case any changes to the SnakeBidsApp need to
```
pvandyken, line 126: The list of returnables might read better at the end  of the docstring, after the context has been fully set?

snakebids/app.py
```diff
@@ -111,6 +113,32 @@ class SnakeBidsApp:
...
+              come from mutating it.
+            - A ``SnakeBidsApp``, which will be used to call Snakemake. Note
```
pvandyken, line 128: I don't know if it's worth highlight the intricacies of returning `SnakeBidsApp`, as it's not something I would even recommend doing unless you have good reason (in which case you should understand the intricacies anyway??). The cost of keeping is weighing down the docs with details most people don't need to know. Perhaps alternatively, we just have a parenthetical: if a new instance of `SnakeBidsApp` is returned it will replace the old instance.

And for Nothing, I might change the tone to: any changes to `SnakeBidsApp` will persist in the workflow.

snakebids/app.py
```diff
@@ -111,6 +113,32 @@ class SnakeBidsApp:
...
+              that in this case, any processing of CLI arguments and
+              configuration must already be handled, so it is recommended to
+              use ``copy.deepcopy`` to copy the original ``SnakeBidsApp``.
+
+        Plugins may perform any arbitrary side effects, including validation,
```
pvandyken, line 133: Maybe add `updates to config` to the list, since that's really the main mutation one might want to perform

snakebids/tests/test_app.py
```diff
@@ -117,12 +117,19 @@ def test_runs_in_correct_mode(
...
+
+        app.add_plugins([plugin])
+
         try:
             app.run_snakemake()
         except SystemExit as e:
             print("System exited prematurely")
             print(e)
 
+        assert app.foo == "bar"
```
pvandyken, line 131: My preference is generally for task-specific unit tests; can we break this out?

pvandyken (changes requested):
The implementation looks good, just left some comments on documentation and the test

pvandyken (approved):

kaitj (approved):

PR #267, opened 2023-03-26
Major updates to documentation

pvandyken:
- Switch theme to furo
- Use snakebids 1.0 syntax for BidsDataset
- Update code blocks with line numbers and highlighting

Not quite finished yet, but I finished drafting the tutorial, so proofreads of that page should be appropriate if your able. Be sure to access the previous page via the actions console below to see the ui updates.


pvandyken:
For the tutorial, I have still have to go over all the code blocks to make sure they are consistent, and check the line numbering and highlighting

pvandyken:
Okay, I'm calling this ready. Documentation is a bit of a black hole for work, so it definitely could be improved still, but I think this is a good stopping point.

## Changes:

* Use literalincludes almost exclusively within tutorial to make
  maintenance easier
* Update non-tutorial documentation files to 1.0 syntax
* Add more links to attributes like zip_lists and path
* Update typing annotations within source code for more consistency, and
  to take advantage of auto-documentation.
* Add instructions to download example dataset
* Various formatting fixes

pvandyken:
Also updated the authors field within pyproject.toml

docs/bids_app/config.md
```diff
@@ -61,24 +65,23 @@ pybids_inputs:
 
 ### `pybids_db_dir`
 
-PyBIDS allows for the use of a cached layout to be used in order to reduce the time required to index a BIDS dataset. A path (if provided) to save the ``PyBIDS`` layout. If ``None`` or ``''`` is provided, the layout is not saved / used. The path provided must be absolute, otherwise the database will not be used. Note, this is a variable used for an opt-in feature and must first be uncommented in the ``snakebids.yml`` file.
+PyBIDS allows for the use of a cached layout to be used in order to reduce the time required to index a BIDS dataset. A path (if provided) to save the *pybids* [layout](#bids.layout.BIDSLayout). If `None` or `''` is provided, the layout is not saved or used. The path provided must be absolute, otherwise the database will not be used.
```
kaitj, line 68: Noted that this likely related to #256 and will be opt-out by 1.0 (unless this has already been implemented and I forgot?). Is the intention to have this PR merged in following review or to hold off until closer to 1.0 release? I believe if we merge this in, these changes will be live in the `latest` version on ReadTheDocs. 

We may want to change the settings around a bit within ReadTheDocs to display the doc reflective of the latest release (`stable`) by default, with option to navigate to latest / main (which would show what is currently living in the `main` branch.

kaitj (approved):
Looks good to me, aside from the one comment.

Thanks for overhauling the documentation! :clap:

docs/bids_app/config.md
pvandyken: Okay, so the confusion here was that the deleted line seemed to refer specifically to the boilerplate app made by `snakebids create`. Users just manually incorporating the database wouldn't be able to uncomment a line.

Now that I look at this, there's a few things to think about here:
* The variable name here is different than the parameter in `generate_inputs` (`pybids_database_dir)`. It's ALSO different from the cli variable (`--pybidsdb-dir`). All of these, in turn, are different from the pybids parameter (`database_path`). I find this situation super confusing.
* Even in the boilerplate app, does the commenting of the config line even do anything if `--pybidsdb-dir` is provided? I remember we had a whole discussion on how to set it up, but the reasoning is not abundantly clear to me any more.
* A general problem I'm now seeing with this config section is it suggests the config variables are special for their downstream effects. This is not the case. They are only special because they are the variable names `SnakeBidsApp` manipulates. But downstream they do nothing, they need to be manually passed to `generate_inputs`. The section stems from the early days where the config was directly passed to `generate_inputs`, and so the structure was super important. But now it's much more a matter of convention

I have a few other thoughts in this vein, especially thinking about plugins and how they might change the terrain. It would be worth have a discussion about this.

pvandyken: I'm inclined to leave the config section as is here, as I'd like to get this merged. It's not super obvious to me how to improve it at this point

docs/tutorial/tutorial.md
```diff
@@ -231,220 +305,205 @@ The Snakefile with the output filename replaced (in both rules) is below:
...
+:end-at: params
+:lineno-match:
+:emphasize-lines: 3
 ```
 
-Step 7: using input wildcards:
-------------------------------
 
-{attr}`BidsDataset.path <snakebids.BidsDataset.path>` already grants us a lot of flexibility, but we can still do more! Next we'll look at {attr}`BidsDataset.wildcards <snakebids.BidsDataset.wildcards>`. This is a dict that maps our input names to the brace-wrapped `{wildcards}` we specified in `pybids_config`. If you printed this value in our test workflow, it would look like this:
+## Step 7: using input wildcards:
```
tkkuehn, line 404: Existing typo: I'd remove the colon at the end of this heading

tkkuehn (approved):
A couple of minor typos (I don't think I need to review it again) but this overall looks great to me, thanks for handling this!

PR #271, opened 2023-04-07
Remove add_plugins method from SnakeBidsApp

pvandyken:
Wanted to sneak this in before the release so we don't need to do a breaking change later. 

Plugins can be trivially added via the constructor, or via SnakeBidsApp.plugins.append, so the method is redundant. Documentation and tests updated to reflect change

### DIFF
--- a/snakebids/app.py
+++ b/snakebids/app.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 import argparse
 import logging
 import sys
-from collections.abc import Iterable
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional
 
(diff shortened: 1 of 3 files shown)


snakebids/app.py
```diff
@@ -97,28 +96,8 @@ class SnakeBidsApp:
...
-    )
-    args: Optional[SnakebidsArgs] = None
-    plugins: list[Callable[[SnakeBidsApp], None | SnakeBidsApp]] = attr.Factory(list)
-
-    def add_plugins(
-        self, plugins: Iterable[Callable[[SnakeBidsApp], None | SnakeBidsApp]]
-    ) -> None:
-        """Supply list of methods to be called after CLI parsing.
+    plugins
+        Supply list of methods to be called after CLI parsing.
```
tkkuehn, line 100: Nitpick: Should get rid of "Supply" as the first word here, because we're now just referring to the attribute

tkkuehn (approved):
One tiny documentation nitpick but otherwise this looks good, thanks

PR #273, opened 2023-04-08
Implement `BidsComponent.expand()`

### LINKED ISSUE #201: `BidsComponent.expand()`
## Problem

Currently, users use `BidsComponents` within the snakemake `expand` function roughly as follows (using the most recent syntax):

```py
expand(
	rules.some_rule.output.some_output,
	my_entity=["foo", "bar"],
	**inputs['my_component'].entities
)
```

Or using zip lists:

```py
expand(
	expand(
		rules.some_rule.ouput.some_output,
		zip,
		allow_missing=True,
		**inputs['my_component'].zip_lists,
	),
	my_entity=["foo", "bar"],
)
```

While functional and transparent, this approach engenders two problems:

1. There's a confusing distinction between `entities` and `zip_lists`. Although `entities`, and its use within `expand`, is much simpler to understand, `zip_lists` is safer and should generally be preferred.
2. As shown in the example above, the use of extra `my_entity=["foo", "bar", ...]` is not straightforward when using the zip list approach. The easiest way is to run `expand` twice, first over the zip list, and then over the custom entities.

Together, these two points increase the cognitive burden on users and the amount of code needed to achieve the typical use-case.

## Proposed Solution

A new `expand` method will be added to `BidsComponent` with the following signature:

```
BidsComponent.expand(paths: PathOrStr | list[PathOrStr], **custom_entities: list[str])
```

In typical use, it will simply call the snakemake expand function over the component entities, using the zip-list method. If `custom_entities` are provided, the two-step expand procedure proposed in problem #&#8203;2 above will be applied. Following the change, both above examples would look like this:

```
inputs['my_component'].expand(
  rules.some_rule.output.some_output),
  my_entity=["foo", "bar"],
)
```

This syntax is easy to understand, concise, and should work in the vast majority of circumstances.

pvandyken:
Resolves #201. Depends on #272 for completion.

Implements the `expand` function on `BidsComponent`, allowing for easy expansion across paths using only the entity-values found in the filesystem. Supports additional wildcards (which get added in standard `product` fashion), multiple paths (as per snakemake `expand`), and, if no paths are given, it expands over the component path, thus re-creating the original files used to form the component.

### Waiting on

- [x] #272


docs/tutorial/step7/Snakefile
```diff
@@ -10,15 +10,14 @@ inputs = generate_inputs(
 
 rule all:
     input:
-        expand(
+        inputs['bold'].expand(
```
tkkuehn, line 13: It may be worth adding `allow_missing=False` to this

docs/bids_app/workflow.md
```diff
@@ -34,21 +34,24 @@ The [](#snakebids.bids) function generates a properly-formatted BIDS filename wi
 
 The {attr}`~snakebids.BidsComponent.path` member of {class}`~snakebids.BidsComponent` is generated by snakebids and contains a list of matched files for every input type. Often, the first rule to be invoked will use one or more entries in `inputs.path` as the input file specification.
 
-The {attr}`~snakebids.BidsComponent.zip_lists` member of {class}`~snakebids.BidsComponent` is used with [`bids()`](#snakebids.bids) and [`expand()`](#snakefiles_expand) to fill the wildcards with corresponding values from input files. The usage pattern is as follows:
+The {meth}`~snakebids.BidsComponent.expand` method of {class}`~snakebids.BidsComponent` is used expand over files using only the entity-values found in your dataset. A usage pattern is as follows:
```
tkkuehn, line 37: Typo: `used expand over` -> `used to expand over`

tkkuehn (approved):
A couple of small comments about the docs, but this looks pretty much good to go.

docs/tutorial/step7/Snakefile
pvandyken: Do you agree with my judgement to make `allow_missing` default to True in general? It's consistent with the way I usually wish `expand` works, but I expected there might be some contention here

tkkuehn: I think I would have left it consistent with the Snakemake default if I had written it, but I'm okay with it either way. I think this way, we get shorter Snakefiles and fewer errors from calls we know we want at the cost of potentially confusing inconsistency with the Snakemake API and the risk of missing error messages we might want

pvandyken: That seems a fair analysis. I'm gonna repost this to a main thread comment, I'd like to get broader consensus here

pvandyken:
One of the decisions I made was to make `allow_missing` default to `True`, as opposed to snakemake's default of `False`. This is consistent with the behaviour I usually expect from `expand`: I do partial expands probably more frequently than full, final expands. @tkkuehn gave a good breakdown of the pros/cons:

> I think I would have left it consistent with the Snakemake default if I had written it, but I'm okay with it either way. I think this way, we get shorter Snakefiles and fewer errors from calls we know we want (due to forgetting the allow_missing=True at the cost of potentially confusing inconsistency with the Snakemake API and the risk of missing error messages we might want

One thing to add is to consider the error messages you might get otherwise. E.g., if you accidentally fail to expand over all the wildcards in a final rule, you'll get a snakemake error saying a rule with wildcards can't be the target rule. If the accidental partial expansion is in the middle of the workflow, I believe you'll get an error saying that the wildcards in the output do not match the wildcards in the input. Will those errors clue someone in easily enough that the problem was they forgot to expand over all the wildcards? (Notably, I don't think the snakemake errors say exactly what wildcards are causing problems). 

Wanted to get consensus on this point before continuing. @akhanf, @kaitj ?

kaitj (approved):
This looks good to me with a couple of minor comments.

kaitj:
> One of the decisions I made was to make `allow_missing` default to `True`, as opposed to snakemake's default of `False`. This is consistent with the behaviour I usually expect from `expand`: I do partial expands probably more frequently than full, final expands. @tkkuehn gave a good breakdown of the pros/cons:
> 
> > I think I would have left it consistent with the Snakemake default if I had written it, but I'm okay with it either way. I think this way, we get shorter Snakefiles and fewer errors from calls we know we want (due to forgetting the allow_missing=True at the cost of potentially confusing inconsistency with the Snakemake API and the risk of missing error messages we might want
> 
> One thing to add is to consider the error messages you might get otherwise. E.g., if you accidentally fail to expand over all the wildcards in a final rule, you'll get a snakemake error saying a rule with wildcards can't be the target rule. If the accidental partial expansion is in the middle of the workflow, I believe you'll get an error saying that the wildcards in the output do not match the wildcards in the input. Will those errors clue someone in easily enough that the problem was they forgot to expand over all the wildcards? (Notably, I don't think the snakemake errors say exactly what wildcards are causing problems).
> 
> Wanted to get consensus on this point before continuing. @akhanf, @kaitj ?

I'm not opposed to this change and think I more or less am with @tkkuehn on this. My (very minor concern) as may have been noted in the review comments left and @tkkuehn pointed out is the "confusing inconsistency".

docs/tutorial/tutorial.md
```diff
@@ -422,17 +422,7 @@ with this:
...
-```py
-inputs['bold'].entities == {
-    'subject': ['001'],
-    'task': ['rest'],
-    'run': ['1', '2'],
-}
-```
-
-This is a useful dict to use with [`expand()`][expand_func] in a target rule, letting us avoid having to specify e.g. the run numbers or task names in the config, and rely on pybids to determine these:
+Finally, we can use our {class}`BidsComponents <snakebids.BidsComponent>` to easily expand over the entity values found in our dataset using {meth}`BidsComponent.expand() <snakebids.BidsComponent.expand>`. This method gets used instead of the snakemake [`expand()`][expand_func] function:
```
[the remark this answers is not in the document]
pvandyken: Hmm, but the expand function is mentioned in previous parts of the tutorial. Should we not maintain the continuity?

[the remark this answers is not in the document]
kaitj: Ahh that is true. If we introduced the **snakemake** expand earlier, this comment can be ignored.

docs/bids_app/workflow.md
```diff
@@ -41,10 +41,11 @@ inputs["bold"].expand(
     bids(
         root="results",
         datatype="func",
-        suffix="bold.nii.gz",
+        suffix="func.shape.gii",
+        hemi="{hemi}",
         **inputs.wildcards["bold"]
     ),
-    myparam=config["myparam_vals"],
+    myparam=config["hemi"],
```
kaitj, line 48: Is `myparam` here suppose to substitute for the `hemi`, so is it to expand over another entity altogether? I think I am use to the snakemake expand and still having a bit trouble understanding how the `myparam` would fit in this call. 

Previously, with the snakemake expand, I would have set something similar to:
```
expand(
  bids(
     ...
     hemi="{hemi}",
  ),
  hemi=config["hemi"]
)
```

kaitj (approved):
Looks good to me with the one minor point below for clarification (which may be due to a lack of :coffee:).

docs/bids_app/workflow.md
pvandyken: No, I'm just sloppy... `myparam` should be `hemi`

PR #275, opened 2023-04-08
Switch default of use_bids_inputs to True

pvandyken:
I thought it was about time we made this switch.

In addition to switching the default, I'd like to deprecate manually setting `use_bids_inputs=True` to attempt to cleanse code of this line. Perhaps we can even remove it completely one day.

One thing I did that I'm not sure any more is helpful is attempt to detect people using `BidsDataset` as if it were `BidsDatasetDict`. If a key from `BidsDatasetDict` is given to `BidsDataset`, and that key is not present in the dataset as a component, an informative error message will be printed. The problem is that the legacy code that might benefit from such a warning are also probably still using the old `config.update(generate_inputs(...))` style. In this case, all the properties from `BidsDataset` will be copied from out of the dataset into the config object, and we lose the opportunity to give the warning. So I'm not sure it's worth even including this warning, as I'd still need to write tests for it, and I doubt anyone will ever see it. Thoughts?

### DIFF
--- a/docs/bids_app/workflow.md
+++ b/docs/bids_app/workflow.md
@@ -20,7 +20,6 @@ inputs = snakebids.generate_inputs(
     derivatives=config.get("derivatives"),
     participant_label=config.get("participant_label"),
     exclude_participant_label=config.get("exclude_participant_label"),
-    use_bids_inputs=True,
 )
 
 ```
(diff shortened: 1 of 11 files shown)


tkkuehn (approved):

tkkuehn:
I think the warning when someone access `BidsDataset["input_path"]` in friends is worth keeping, given that the work to put it in is already done, and it has the potential to help people catch legacy-style use of `BidsDataset` as they transition to the current style.

pvandyken:
Sounds good. I added some tests, if you care to re-review

kaitj (approved):

PR #276, opened 2023-04-11
Create migration guide for v0.8

pvandyken:
Went ahead and wrote up a migration guide, and updated the previous guide, given the bevy of changes. 

(Note that some doc links are broken until #273 is merged)


docs/migration/0.7_to_0.8.md
```diff
@@ -0,0 +1,50 @@
...
+
+# should now use
+inputs["t1w"].wildcards
+inputs["t1w"].entities
+```
+
+{#0_8_migration_expand_func}
+## New {meth}`~snakebids.BidsComponent.expand` method
+
+V0.8 features a new {meth}`~snakebids.BidsComponent.expand` method on {class}`~snakebids.BidsComponent`. This method automatically ensures only entity-values actually contained in your dataset are used when expanding over a path. It supports the addition of extra wildcards, and can expand over the component {attr}`~snakebids.BidsComponent.path` or any number of provided paths. It should generally be preferred over snakemake's [`expand()`](#snakefiles_expand) when {class}`BidsComponents <snakebids.BidsComponent>` are involved, due to the increased safety and easy of use.
```
kaitj, line 48: "easy of use" -> "ease of use"

kaitj (approved):
LGTM. I wonder about renaming the "0.5_to_0.6.md" to "0.5_to_0.8.md" for consistency, but also this isn't a big deal.

If it's not already done, we might benefit from including tagged releases for readthedocs, such that we can point to previous migration guides (e.g. 0.5 -> 0.6 -> 0.8) to bring the user up-to-speed instead of updating (a potentially increasing number of) guides to the latest version.

pvandyken:
> If it's not already done, we might benefit from including tagged releases for readthedocs, such that we can point to previous migration guides (e.g. 0.5 -> 0.6 -> 0.8) to bring the user up-to-speed instead of updating (a potentially increasing number of) guides to the latest version.

We do have tagged releases... My plan is to have one per minor version. I don't think we'll need migration guide for every version update, just for the breaks and deprecations before version 1.0. So yeah, if we ever find ourselves wanting to write another, I would say just lose one of the previous guides so that we keep the number limited to 2

tkkuehn (approved):
There's one addition that I think would be helpful if you have time but this otherwise looks good, thanks!

PR #278, opened 2023-04-19
Switch from pylint/flake8 to ruff

pvandyken:
Discovered this [library](https://beta.ruff.rs/docs/) recently. It replicates all the functionality of pretty much every linter I know of, all way faster. It solves the dependency woes of `flake8`, and it's a little easier to manage, including some special modes for automatically [adding and removing noqa tags](https://beta.ruff.rs/docs/configuration/#automatic-noqa-management). I think it's definitely worth switching over


docs/conf.py
```diff
@@ -1,4 +1,3 @@
-# pylint: disable=all
 # flake8: noqa
```
kaitj, line 1: If switching to ruff, this line can probably be removed.

snakebids/tests/mock/config.py
```diff
@@ -1,5 +1,4 @@
 # flake8: noqa
```
kaitj, line 1: Another one that can be removed?

kaitj (commented):
Took a quick glance through and it looks good to me - just want to take a little more time to take a look at ruff. There are a few things leftover from flake8 here that I think can be removed if the switch is made?

snakebids/tests/mock/config.py
pvandyken: Or switch to `# ruff: noqa`. Ruff does respect the flake8 codes though, but happy to switch

kaitj: Ahh - I didn't know that. I think to avoid any future confusion, might be good to switch, but also open to leaving it as flake8

snakebids/cli.py
```diff
@@ -26,20 +25,18 @@ def __call__(self, parser, namespace, values, option_string=None):
 
         for value in values:
             # split it into key and value
-            key, value = value.split("=")
+            key, value = value.split("=")  # noqa: PLW2901
```
tkkuehn, line 28: We could dodge the `noqa` by changing this to something like `for key, value in [value.split('=') for value in values]:`

snakebids/core/input_generation.py
```diff
@@ -163,19 +160,19 @@ def generate_inputs(
         ├── README
         └── sub-control01
             ├── anat
-            │   ├── sub-control01_T1w.json
-            │   ├── sub-control01_T1w.nii.gz
-            │   ├── sub-control01_T2w.json
-            │   └── sub-control01_T2w.nii.gz
+            │   ├── sub-control01_T1w.json
```
tkkuehn, line 163: Probably unrelated to the PR but do you know why this is all showing up in the diff?

snakebids/tests/mock/config.py
tkkuehn: My preference would be to switch

tkkuehn (approved):
Looks good to me -- I hadn't heard of this (and wouldn't have considered pylint's speed an issue for Snakebids at this point), but if you're satisfied that it covers pylint's checks, the speed is a minor benefit and we get rid of a lot of pylint disables, which seems like a win.

snakebids/core/input_generation.py
pvandyken: I changed a bunch of ambiguous special white-space characters into ordinary spaces

pvandyken:
Also, in VScode, ruff has a nice extension with full hover support, including over `noqa` codes. So if you hover over a `noqa`, it gives a nice description of the code's meaning

PR #282, opened 2023-04-25
Final documentation fixes before release

pvandyken:
One last PR! I realized the `generate_inputs` docstring was never updated, so I wanted to fix it.

While at it, I wanted to make sure we had working links for all our `TypeAliases`, especially `ZipList`. Unfortunately, autodoc doesn't support this well natively, but I was able to find a work-around (after an unfortunately long rabbit hole). I documented the hack in the comments.

Most of the API is now visible on the website, however, I split it into two pages: one with the "main" stuff and one with the "internal" stuff.


tkkuehn (approved):
Thanks for this -- the type alias workaround in particular seems like it would have been a pain to figure out, but well worth it. The new `generate_inputs` docs look great, too.

PR #283, opened 2023-04-27
Restrict pybids bug check to python 3.7

pvandyken:
Now merging through the proper channels!

The underlying [issue](https://github.com/bids-standard/pybids/issues/788) does appear to be resolved, so I restricted the failing test to py37, where it's still relevant (because we can't install the latest version of pybids on py37)

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1713,14 +1713,14 @@ files = [
 
 [[package]]
 name = "platformdirs"
-version = "3.4.0"
+version = "3.5.0"
 description = "A small Python package for determining appropriate platform-specific dirs, e.g. a \"user data dir\"."
 category = "dev"
 optional = false
 python-versions = ">=3.7"
 files = [
-    {file = "platformdirs-3.4.0-py3-none-any.whl", hash = "sha256:01437886022decaf285d8972f9526397bfae2ac55480ed372ed6d9eca048870a"},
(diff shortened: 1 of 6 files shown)


tkkuehn (approved):

pvandyken:
Just a flaky hypothesis test (due to test timings). A re-run usually solves it. Not ideal, but seems to be the nature of the beast

kaitj:
Lgtm, though now the docs are failing to build (dependency conflicts?)

On the failure, is there a way to set the maximum runtime for the tests? Highly doubt it would happen (knock on wood), but could it ever happen where re-runs will fail for at least 1 python version causing it the checks to repeatedly fail. Not a big deal in any case if the re-run will solve the issue and don't want to limit the testing either.

kaitj (approved):

pvandyken:
> Highly doubt it would happen (knock on wood)

Too late, you've jinxed it! I've increased the deadline for the two flaky tests.

pvandyken:
On the strictly maintenance front, I added a `poe fix` command that applies `ruff` autofixes

pvandyken:
Okay, this turned into sort of an omnibus issue, but I think everything is sorted. I don't know why the tests were flaking so much... maybe the git actions runner was hanging?

The only other thing I did was fix some of my tests which were calling `strategy.example` directly. Turns out you can just set the `max_example` parameter in `setting` to 1, so that's what I did

PR #285, opened 2023-05-01
Implement extension support

### LINKED ISSUE #235: Support `extension` in `bids()` function
## The problem

From a user perspective, `extension` is somewhat redundant, as they can just use `suffix`. However, as of #229, extensions can be indexed and "wildcarded" by snakebids, so we need to support them within `bids()`

pvandyken:
~This PR is a draft because I'm waiting on the latest pybids release (any time now) which corrects some upstream bugs preventing successful merging. But the code should be good for review.~

Here we allow `bids()` to handle extension. Resolves #235. Not useful directly, but necessary if users want to use extension as a wildcard in `pybids_inputs`.

Along the way:
- Standardize filtering of path-safe entity-lists into function
- Implement entity white- and blacklist support throughout strategies
- Remove the arbitrary addition of '.foo' in suffix-laden paths in
  make_bids_path, pushing entity-suffix pairing into the entity-list
  filtering function## Proposed changes
- Update the regex for suffix and extension to match with latest pybids

The tests introduced in this PR will generally not work on py37 because the latest pybids is `>=py38`. So I just disabled the testing of `suffix` and `extension` entities on py37. The actual implementation isn't really affected: users probably won't notice the bug even on old versions of pybids.


pvandyken:
Okay, assuming the tests pass, this is ready for review. This ended up being a little bit more complicated than expected, and there's a bunch of improvements to the testing code. The details are summarized in the commit messages but a few notes here:

* I bumped pyright to the latest version. It's starting to complain about the dicts we pass in the test code to `MultiSelectDict` for some reason, I presume this is a regression, because it used to infer the type just fine. BUT, rather than fight it or add a bunch of ignores, I'm taking advantage of one of the latest features, which is to respect converters in attrs classes. So I actually just removed all the `MultiSelectDict` type conversions because `BidsComponent` already does this via attrs, and now pyright can recognize this.
* I completely refactored `_parse_bids_path`. The original implementation wasn't able to handle `{suffix}{extension}` because the two wildcards were directly adjacent, and this caused problems. In hindsight, I could have made some fairly simple modifications to make this work, but I didn't realize this at first and ended up taking a different approach. The new version is 25-50% faster, so I left it in place even though it's a bit more complicated. 
* The testing strategies for Component/Dataset now generally support adding additional, fixed entities to the `path` of each component, making our examples ever more realistic

kaitj (commented):
First pass this mostly looks good. Were the `out` and `.testmondata*` files added in during local tests?

pvandyken:
> First pass this mostly looks good. Were the `out` and `.testmondata*` files added in during local tests?

Woops, must have got included in an errant `git add`

snakebids/core/input_generation.py
```diff
@@ -485,23 +491,48 @@ def _parse_bids_path(path: str, entities: Iterable[str]) -> tuple[str, dict[str,
...
+    # If path is relative, we need to get a slash in front of it to ensure parsing works
+    # correctly. So prepend "./" or ".\" and run function again, then strip before
+    # returning
+    if not os.path.isabs(path) and not get_first_dir(path) == ".":
+        path_, wildcard_values = _parse_bids_path(os.path.join(".", path), entities)
+        return str(Path(path_)), wildcard_values
+
+    entities = list(entities)
+
+    def sort_by_match(match: tuple[BidsEntity, re.Match[str]]):
```
tkkuehn, line 503: Generally I prefer to define utility functions like this at the top level if they don't use any function-scoped variables (even if they're only used within the context of the function). But I don't insist, especially given that this is probably mostly not a lambda to prevent pyright from complaining

snakebids/tests/strategies.py
```diff
@@ -23,32 +24,108 @@
...
+@st.composite
+def bids_path(  # noqa: PLR0913
+    draw: st.DrawFn,
+    *,
+    root: PathLike[str] | str | None = None,
+    entities: Iterable[BidsEntity | str] | None = None,
+    blacklist_entities: Optional[Container[BidsEntity | str]] = None,
+    whitelist_entities: Optional[Container[BidsEntity | str]] = None,
+    extra_entities: bool = True,
+):
```
tkkuehn, line 80: Optional: would be nice to annotate the return type here

tkkuehn (approved):
A couple of minor stylistic things, but this looks great overall and I won't need to look at it again. Thanks!

snakebids/core/input_generation.py
pvandyken: I think in a previous version I was calling this function multiple times, rather than the current one-off. Pyright is able to infer the arguments in a `sorted` key tuple, so I switched it to a lambda

pvandyken:
@kaitj, let me know if you have any comments/suggestions

kaitj (approved):

PR #290, opened 2023-05-04
Ignore deadlines for hypothesis tests on pr

pvandyken:
Been having a lot of issues with flaky tests lately, entirely due to timing inconsistencies. The failing tests are not complicated, there's not particular reason for long runtimes, and it probably has more to do with the github actions runner being slow (here's another source corroborating this https://datascience.blog.wzb.eu/2022/03/04/continuous-integration-testing-with-github-actions-using-tox-and-hypothesis/). So I'd like to just disable test deadlines altogether on for PRs: playing whack-a-mole one test a time is too much work.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -117,6 +117,8 @@ jobs:
     - name: Install library
       run: poetry install --no-interaction
     - name: Test with pytest
+      env:
+        HYPOTHESIS_PROFILE: pr
       run: |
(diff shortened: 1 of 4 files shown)


kaitj:
I think I am okay with this as long as I am understanding how the `deadline` works - that is setting the max time allotted time that each test function call is allowed to exceed. 

Is there anyway to set the maximum amount of time for the entire test suite if this isn’t already set? I think by default actions can run for up to 6 hours iirc.

tkkuehn (approved):

kaitj (approved):

PR #291, opened 2023-05-09
Bids-validator plugin

kaitj:
## Proposed changes

A second attempt at implementing bids validation (see #222). This time, implementing as a snakebids plugin (resuing some of the code from the previous attempt). The plugin tries to perform validation using the bids-validator unless skipped by the user (`--skip_bids_validation`). If the validation fails, an `InvalidBidsError` is thrown, otherwise validation is performed by default using pybids with a warning informing the user.

As noted in #222:
> Drawing some inspiration from fmriprep, a temporary config file is generated for the system call in (_validate_bids_dir) which includes files to ignore during this validation. Was also wondering if we want to limit the validation to the subject data themselves rather than a full dataset (e.g. ignore participants.tsv, dataset_description.json)?

## Checklist

## Notes


snakebids/core/input_generation.py
```diff
@@ -283,12 +291,13 @@ def _all_custom_paths(config: InputsConfig):
     return all(comp.get("custom_path") for comp in config.values())
 
 
-def _gen_bids_layout(
+def _gen_bids_layout(  # noqa: PLR0913
```
pvandyken, line 294: PLR0913 can be resolved by making the arguments kw-only. I think this is a good approach

snakebids/plugins/validate.py
```diff
@@ -0,0 +1,58 @@
...
+import logging
+import subprocess
+import tempfile
+
+from snakebids.app import SnakeBidsApp
+
+_logger = logging.getLogger(__name__)
+
+
+class InvalidBidsError(Exception):
```
pvandyken, line 11: It would be nice to have some base snakebids exception to inherit from. Either `SnakebidsException` or `SnakebidsPluginException`?

snakebids/plugins/validate.py
```diff
@@ -0,0 +1,58 @@
...
+    not found.
+
+    Parameters
+    ----------
+    app
+        Snakebids application to be run
+    """
+
+    def __init__(self, app: SnakeBidsApp) -> None:
+        self.app = app
```
pvandyken, line 27: The app here should be moved to `__call__`. `__init__` can be used for any user-specified configuration

snakebids/plugins/validate.py
```diff
@@ -0,0 +1,58 @@
...
+            with tempfile.NamedTemporaryFile(mode="w+", suffix=".json") as temp:
+                temp.write(json.dumps(validator_config_dict))
+                temp.flush()
+
+            subprocess.check_call(
+                ["bids-validator", self.app.config["bids_dirs"], "-c", temp.name]
+            )
+
+            # If successfully bids-validation performed, skip pybids validation
+            self.app.config["bids_validator_success"] = True
```
pvandyken, line 46: Skipping pybids isn't strictly necessary, since pybids doesn't actually do validations (in the latest versions, this is literally true: `validate=True` just causes it to ignore unrecognized files). I also don't know off-hand how pybids handles `.bidsignore` files. If it doesn't read that, keeping pybids validation on is even more important

snakebids/project_template/{{cookiecutter.app_name}}/config/snakebids.yml
```diff
@@ -81,11 +81,20 @@ parse_args:
...
+
+ # custom command-line parameters can then be added, these will get added to the config and also accessible to plugins
+ # below are an examples for plugin and custom parameters (e.g. config['smoothing_fwhm'])
+  --skip_bids_validation:
+    help: 'Skip validation of BIDS dataset. BIDS validation is performed by 
+          default using the bids-validator (if installed) or with the pybids
+          validator implementation (if bids-validator is not installed). 
+          (default: %(default)s)'
+    action: "store_true"
+    default: False
```
pvandyken, line 93: Can we have `dest: plugins.validator.skip` here. Good to start enforcing namespacing right from the get-go

snakebids/plugins/validate.py
pvandyken, line 46: Here again, lets use a namespaced location: `plugins.validator.success`. I think just using keys with dots is the way to go here: it accomplishes namespacing just as well as a nested dict, it's shorter to right, and it saves complexity, as we don't need to normalize the `dest` tag from the cli into a nested dict.

snakebids/tests/test_validate_plugin.py
```diff
@@ -0,0 +1,85 @@
...
+        validator()
+
+        # Check validation failure
+        assert not app.config["bids_validator_success"]
+
+        # Check logger message (should only have 1 warning message)
+        assert len(caplog.records) == 1
+        assert caplog.records[0].levelname == "WARNING"
+        assert (
+            "Missing bids-validator installation - falling back to pybids validation."
```
pvandyken, line 68: I think my preference here is to just check if a relevant substring is `in` the warning: enough to make sure we got the right warning message, but not so much that the test immediately breaks if we slightly alter the warning (obviously a total rewrite of the warning will always break the test)

pvandyken (changes requested):

snakebids/plugins/validate.py
```diff
@@ -0,0 +1,58 @@
...
+
+        # If the bids-validator call can't be made
+        except FileNotFoundError:
```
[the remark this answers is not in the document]
kaitj: This part was borrowed from fmriprep and was tested a couple months ago during the first attempt at implementing the validator. That said, I realized this needs to be slightly modified

snakebids/core/input_generation.py
```diff
@@ -283,12 +291,13 @@ def _all_custom_paths(config: InputsConfig):
...
     bids_dir: Path | str,
     derivatives: Path | str | bool,
     pybids_database_dir: Path | str | None,
     pybids_reset_database: bool,
     pybids_config: Path | str | None = None,
+    validate: bool = True,
```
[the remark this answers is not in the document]
kaitj: Yes, I agree - meant to do this, but forgot to do it before pushing.

snakebids/plugins/validate.py
kaitj: From what I could tell from a few threads that are a couple of years old, the `.bidsignore` doesn't do anything in 
pybids. 

Trying to think if there are any situations where pybids validation would fail where we would not want it to and at the moment I can't come up with any. With that in mind, I see no reason not to have pybids validation on. That being said, does setting `validate=True` (e.g. keeping pybids validation on) not conflict with the suggestion below for setting the default to be `validate=False`?

docs/running_snakebids/overview.md
```diff
@@ -15,7 +15,9 @@ Note that if any rules in the Snakebids workflow use Singularity containers, spe
 Indexing of large datasets can be a time-consuming process. Snakebids, through `PyBIDS` has the ability to create or leverage an existing database, requiring indexing of datasets to be only performed when user chooses to do so (usually if the dataset has changed)! Note, this feature is **opt-in**, meaning it is not used unless the associated config variables are used. To opt-in:
 
 1. Uncomment the lines in `snakebids.yml` containing `pybids_db_dir` and `pybids_db_reset`.
-1. The variables can be updated directly in this file or through the CLI by using `-pybidsdb-dir {dir}` to specify the database path and `--reset-db` to indicate that the database should be updated. _Note: CLI arguments take precendence if both CLI and config variables are set._
+1. The variables can be updated directly in this file or through the CLI by using `--pybidsdb-dir {dir}` to specify the database path and `--reset-db` to indicate that the database should be updated. _Note: CLI arguments take precendence if both CLI and config variables are set._
```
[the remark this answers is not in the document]
kaitj: Agreed. I think my preference here though might be to update the plugins section of the doc using the validator to: (1) show an updated example of using classes to create plugins and (2) show how to enable the plugin and point users there.

Ultimately, if we have plugins we are creating and distributing with snakebids, is it worth to create a list of available plugins somewhere?

snakebids/plugins/validate.py
kaitj: Hmm, I like the idea of `SnakebidsPluginException` for something like this, especially if with enabling developers to create their own plugins.

snakebids/plugins/validate.py
kaitj: Ahh, I see - had to reread that a couple of times :sweat_smile:. 

So then I think the code currently stands with validate=False throughout should satisfy the first condition. And for the second condition, it would be updating the cookiecutter to:

```py
inputs = generate_inputs(
   ...
   validate=not(config.get("bids_validator_success", False))
```

in which case the pybids validation will be performed regardless of whether the bids-validator can be called so long as the user didn't choose to skip validation.

docs/running_snakebids/overview.md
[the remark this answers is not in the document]
kaitj: Revamped the plugins part of the docs a little, building off of what Tristan had already started in #261. Split up use and creation of plugins into two sections, and also added a couple notes on recommending namespaced keys for the config and making use of the `SnakebidsPluginError`.

docs/bids_app/plugins.md
```diff
@@ -1,32 +1,71 @@
 # Plugins
 
-Plugins are a Snakebids feature that allow you to add arbitrary behaviour to your Snakebids app after CLI arguments are parsed but before Snakemake is invoked. For example, you might add BIDS validation of an input dataset to your app via a plugin, so your app is only run if the input dataset is valid.
+Plugins are a feature in Snakebids that enables you to add custom functionality to your Snakebids application after parsing CLI arguments but before invoking Snakemake. For example, you can use a plugin to perform BIDS validation to your SnakeBids app, which ensures your app is only executed if the input dataset is valid. You can either use those that are distributed with Snakebids (see [Using plugins](#using-plugins)) or create your own plugins (see [Creating plugins](#creating-plugins)).
```
tkkuehn, line 3: SnakeBids -> Snakebids

tkkuehn, line 3: "to perform BIDS validation to your SnakeBids app" -> "to perform BIDS validation of your Snakebids app's input"

docs/bids_app/plugins.md
```diff
@@ -1,32 +1,71 @@
...
 
-A plugin is simply a function that takes a {class}`SnakeBidsApp <snakebids.app.SnakeBidsApp>` as input and returns either a modified {class}`SnakeBidsApp <snakebids.app.SnakeBidsApp>` or `None`. To add one or more plugins to your {class}`SnakeBidsApp <snakebids.app.SnakeBidsApp>`, pass them to the {class}`~snakebids.app.SnakeBidsApp` constructor via the {attr}`~snakebids.app.SnakeBidsApp.plugins` parameter. Your plugin will have access to CLI parameters (after they've been parsed) via their names in {attr}`SnakeBidsApp.config <snakebids.app.SnakeBidsApp.config>`. Any modifications to that config dictionary will be carried forward into the workflow.
+Using plugins
+--------------
```
tkkuehn, line 6: I generally prefer the `##` syntax for subheadings

docs/bids_app/plugins.md
```diff
@@ -1,32 +1,71 @@
...
+To use one or more plugins to your {class}`SnakeBidsApp <snakebids.app.SnakeBidsApp>`, pass them to the {class}`~snakebids.app.SnakeBidsApp` constructor via the {attr}`~snakebids.app.SnakeBidsApp.plugins` parameter. Your plugin will have access to CLI parameters (after they've been parsed) via their names in {attr}`SnakeBidsApp.config <snakebids.app.SnakeBidsApp.config>`. Any modifications to that config dictionary made by the plugin will be carried forward into the workflow.
```
tkkuehn, line 7: use -> add

docs/bids_app/plugins.md
```diff
@@ -1,32 +1,71 @@
...
 SnakeBidsApp(
     "path/to/snakebids/app",
-    plugins=[bids_validate]
+    plugins=[BidsValidator]
 ).run_snakemake()
 ```
 
-You would also want to add some logic to check if the BIDS Validator is installed and pass along the error message, but the point is that a plugin can do anything that can be handled by a Python function.
+Creating plugins
+----------------
```
tkkuehn, line 24: `##` syntax?

docs/bids_app/plugins.md
```diff
@@ -1,32 +1,71 @@
...
+A plugin is a callable class that accepts a {class}`SnakeBidsApp <snakebids.app.SnakeBidsApp>` as input and returns a modified {class}`SnakeBidsApp` or `None`.
+
+As an example, the creation of the bids-validator plugin to run the [BIDS Validator](https://github.com/bids-standard/bids-validator) can be created like so:
```
tkkuehn, line 27: Maybe "As an example, a simplified version of the bids-validator plugin that runs the [BIDS Validator] could be defined as follows:"

docs/bids_app/plugins.md
```diff
@@ -1,32 +1,71 @@
...
+    inheriting from SnakebidsPluginError.
+    """
+```
+
+```{note}
+When adding plugin-specific parmaters to the config dictionary, it is recommended to use namespaced keys (e.g. ``plugin.validator.skip``). This will help ensure plugin-specific parameters do not conflict with other parameters already defined in the dictionary or by other plugins.
+```
+
+
+A plugin can be used to implement any logic that can be handled by a Python function. In the above example, you may also want to add some logic to check if the BIDS Validator is installed and pass along a custom error message if it is not. Created plugins can then be used within a Snakebids workflow, similar to the example provided in [Using plugins](#using-plugins) section.
```
tkkuehn, line 66: Maybe note that prospective users can take a look at the source of the "snakebids.plugins" module for examples?

docs/running_snakebids/overview.md
```diff
@@ -15,7 +15,9 @@ Note that if any rules in the Snakebids workflow use Singularity containers, spe
...
+
+Validation of input BIDS datasets are also enabled by default using the bids-validator plugin. By default, this feature uses the command-line (node.js) version of the [validator](https://www.npmjs.com/package/bids-validator). If this is not found to be installed on the system, the `pybids` version of validation will be performed instead. To opt-out of validation, invoke the `--skip-bids-validation` flag. Details related to using and creating plugins can be found on the [plugins](/bids_app/plugins) page.
```
tkkuehn, line 20: "are -> is" in the first sentence

snakebids/plugins/validator.py
```diff
@@ -0,0 +1,59 @@
...
+                # If the bids-validator call can't be made
+                self.app.config["plugins.validator.success"] = False
+                _logger.warning(
+                    "Missing bids-validator installation - falling back to pybids "
+                    "validation."
+                )
+            # Any other bids-validator error
+            except subprocess.CalledProcessError as err:
+                self.app.config["plugins.validator.success"] = False
+                raise InvalidBidsError from err
```
tkkuehn, line 59: This is optional, but maybe we want to make it configurable whether we raise an exception if the BIDS Validation fails? Catching this error would be kind of a pain (you'd have to do it from `run.py`, and it's possible a user might want to know whether bids validation was successful and change its behaviour accordingly.

snakebids/plugins/validator.py
```diff
@@ -0,0 +1,59 @@
...
+    ----------
+    app
+        Snakebids application to be run
+    """
+
+    def __init__(self) -> None:
+        pass
+
+    def __call__(self, app: SnakeBidsApp) -> None:
+        self.app = app
```
tkkuehn, line 31: Is there a reason to assign `self.app` here?

snakebids/project_template/{{cookiecutter.app_name}}/config/snakebids.yml
```diff
@@ -81,11 +81,21 @@ parse_args:
     default: False
     nargs: '+'
 
- # custom command-line parameters can then be added, these will get added to the config
- # below is an example to override config['bet_frac']
+
+ # custom command-line parameters can then be added, these will get added to the config and also accessible to plugins
+ # below are an examples for plugin and custom parameters (e.g. config['smoothing_fwhm'])
```
tkkuehn, line 86: "below are an examples" -> "below are examples"

snakebids/tests/test_validate_plugin.py
```diff
@@ -0,0 +1,83 @@
...
+    def app(self):
+        return SnakeBidsApp(
+            Path("app"),
+            skip_parse_args=False,
+            snakefile_path=Path("Snakefile"),
+            configfile_path=Path("mock/config.yaml"),
+            config=copy.deepcopy(config),
+        )
+
+    _logger = logging.getLogger(__name__)
```
tkkuehn, line 27: Looks like this might be left over from some debugging?

tkkuehn (changes requested):
Looking good, just a few typos and one suggestion for a potential enhancement to the plugin

snakebids/plugins/validator.py
kaitj: None in particular other than it was already self assigned before and I just moved the code down rather than search and replace all to just `app`. Can make that change later today though.

snakebids/tests/test_validate_plugin.py
kaitj: Ahh I have to double check, this was leftover when I was trying to get the warnings to be asserted probably

docs/bids_app/plugins.md
kaitj: Is this a change we want to make throughout. It would be inconsistent with the rest of the docs otherwise.

tkkuehn: I thought the capitalization we typically go with was "Snakebids". A quick look at the docs seems to support that impression

snakebids/plugins/validator.py
kaitj: My initial thought is to leave it as is as I don’t see a situation where the behaviour changes based on validation pass or failure of the input dataset. Open to changing it though.

docs/bids_app/plugins.md
kaitj: I think I may leave this as is for now. It would be consistent with some of the other pages and also to leave in the anchors. Does anyone know if the `##` also create anchors?

pvandyken: The only reason other pages use `---------` is because they used to be in `.rst`. If by anchors, you mean ids you can link to, yes the `##` syntax does make links.

snakebids/plugins/validator.py
pvandyken: Seems easy enough to add a parameter to the `__init__` of the class. Default it to keeping the exception. It adds another condition to test, but if that's not too much hardship, I see no reason not to make this configurable

kaitj: Fair enough - easy enough to add a test for this too. I've changed this in the latest commit.

tkkuehn (approved):
Yeah it is not at all clear to me what's going on with that workflow failure, I think we can probably just merge and address it in a separate PR if it sticks around

kaitj:
Looks like making a new push fixed that workflow failure.

pvandyken:
@kaitj, are you still working on this one, or am I good to take another look?

kaitj:
> @kaitj, are you still working on this one, or am I good to take another look?

Should be good for another look.

docs/bids_app/plugins.md
```diff
@@ -1,32 +1,69 @@
...
 SnakeBidsApp(
     "path/to/snakebids/app",
-    plugins=[bids_validate]
+    plugins=[BidsValidator]
 ).run_snakemake()
 ```
 
-You would also want to add some logic to check if the BIDS Validator is installed and pass along the error message, but the point is that a plugin can do anything that can be handled by a Python function.
+## Creating plugins
+A plugin is a callable class that accepts a {class}`SnakeBidsApp <snakebids.app.SnakeBidsApp>` as input and returns a modified {class}`SnakeBidsApp` or `None`.
```
pvandyken, line 23: I would clarify `function or callable class`, just for those who aren't as intimate in python and won't realize that those are essentially the same thing

docs/bids_app/plugins.md
```diff
@@ -1,32 +1,69 @@
...
+
+
+class InvalidBidsError(SnakebidsPluginError):
+    """Error raised if input BIDS dataset is invalid, 
+    inheriting from SnakebidsPluginError.
+    """
+```
+
+```{note}
+When adding plugin-specific parmaters to the config dictionary, it is recommended to use namespaced keys (e.g. ``plugin.validator.skip``). This will help ensure plugin-specific parameters do not conflict with other parameters already defined in the dictionary or by other plugins.
```
pvandyken, line 60:
```suggestion
When adding plugin-specific parameters to the config dictionary, it is recommended to use namespaced keys (e.g. ``plugins.validator.skip``). This will help ensure plugin-specific parameters do not conflict with other parameters already defined in the dictionary or by other plugins.
```

snakebids/project_template/{{cookiecutter.app_name}}/workflow/Snakefile
```diff
@@ -13,6 +13,7 @@ inputs = generate_inputs(
     derivatives=config.get("derivatives", None),
     participant_label=config.get("participant_label", None),
     exclude_participant_label=config.get("exclude_participant_label", None),
+    validate=True if "plugins.validator.success" in config else False,
```
pvandyken, line 16:
```suggestion
    validate=config.get("plugins.validator.success", False),
```

snakebids/tests/helpers.py
```diff
@@ -205,7 +205,7 @@ def reindex_dataset(
     if use_custom_paths:
         for comp in config:
             config[comp]["custom_path"] = dataset[comp].path
-    return generate_inputs(root, config)
+    return generate_inputs(root, config, validate=False)
```
pvandyken, line 208: `validate` defaults to `False` already I think? So you don't need this

snakebids/tests/test_generate_inputs.py
```diff
@@ -109,7 +109,7 @@ def test_ambiguous_paths_with_extra_entities_leads_to_error(
         }
 
         with pytest.raises(ConfigError):
-            generate_inputs(root, pybids_inputs)
+            generate_inputs(root, pybids_inputs, validate=False)
```
pvandyken, line 112: Ditto for all these

snakebids/tests/test_validate_plugin.py
```diff
@@ -0,0 +1,96 @@
...
+
+        validator = BidsValidator()
+        validator(app)
+        assert app.config["plugins.validator.success"]
+
+    def test_missing_bids_validator(
+        self, app: SnakeBidsApp, mocker: MockerFixture, caplog: pytest.LogCaptureFixture
+    ):
+        # Test fall back to Pybids validation
+        mocker.patch("subprocess.check_call", side_effect=FileNotFoundError)
```
pvandyken, line 49: As far as I know, `subprocess.check_call` will never actually raise a `FileNotFoundError`... could you confirm?

pvandyken (changes requested):
Apologies again for the delay, I left some comments

snakebids/tests/test_validate_plugin.py
kaitj: > image

From running `subprocess.check_call` manually

pvandyken: Ah, okay great. Apologies for my doubt, haha

docs/running_snakebids/overview.md
```diff
@@ -15,7 +15,9 @@ Note that if any rules in the Snakebids workflow use Singularity containers, spe
 Indexing of large datasets can be a time-consuming process. Snakebids, through `PyBIDS` has the ability to create or leverage an existing database, requiring indexing of datasets to be only performed when user chooses to do so (usually if the dataset has changed)! Note, this feature is **opt-in**, meaning it is not used unless the associated config variables are used. To opt-in:
 
 1. Uncomment the lines in `snakebids.yml` containing `pybids_db_dir` and `pybids_db_reset`.
```
pvandyken, line 17: Sorry for missing this last night, but this step should be removed I think, as it's no longer relevant

pvandyken: And of course, without the first step, probably makes sense to remove step labels and have plain prose

docs/running_snakebids/overview.md
```diff
@@ -15,7 +15,9 @@ Note that if any rules in the Snakebids workflow use Singularity containers, spe
...
-1. The variables can be updated directly in this file or through the CLI by using `-pybidsdb-dir {dir}` to specify the database path and `--reset-db` to indicate that the database should be updated. _Note: CLI arguments take precendence if both CLI and config variables are set._
+1. The variables can be updated directly in this file or through the CLI by using `--pybidsdb-dir {dir}` to specify the database path and `--reset-db` to indicate that the database should be updated. _Note: CLI arguments take precendence if both CLI and config variables are set._
+
+Validation of input BIDS dataset is also enabled by default using the bids-validator plugin. By default, this feature uses the command-line (node.js) version of the [validator](https://www.npmjs.com/package/bids-validator). If this is not found to be installed on the system, the `pybids` version of validation will be performed instead. To opt-out of validation, invoke the `--skip-bids-validation` flag. Details related to using and creating plugins can be found on the [plugins](/bids_app/plugins) page.
```
pvandyken, line 20: Enabled by default seems a bit misleading, it's more correct to say the boilerplate app starts with the validator plugin enabled. But without the plugin we don't do any validation

snakebids/core/input_generation.py
```diff
@@ -139,6 +142,10 @@ def generate_inputs(  # noqa: PLR0913
         :class`BidsDataset`. Setting to True is deprecated as of v0.8, as this is now
         the default behaviour
 
+    validate
+        If False, skips validation of BIDS directory. Otherwise use node.js
```
pvandyken, line 146: The validation done by `generate_inputs` will never touch `node.js`, that's strictly a plugin thing. I'd keep the docstring here focused

pvandyken: i.e. the behaviour of this validate doesn't change when the plugin is used, just the way `config['plugins.validate.success']` or whatever changes

snakebids/plugins/validator.py
```diff
@@ -0,0 +1,58 @@
...
+            try:
+                subprocess.check_call(
+                    ["bids-validator", app.config["bids_dirs"], "-c", temp.name]
+                )
+
+                # If successfully bids-validation performed
+                app.config["plugins.validator.success"] = True
+            except FileNotFoundError:
+                # If the bids-validator call can't be made
+                app.config["plugins.validator.success"] = False
```
pvandyken, line 49: A bit confused here because in the boilerplate app, validate is set to the value of this sucess parameter, so if `FileNotFound`, I don't think it's falling back to pybids.

snakebids/project_template/{{cookiecutter.app_name}}/{{cookiecutter.app_name}}/run.py
```diff
@@ -11,7 +12,10 @@ def get_parser():
 
 
 def main():
-    app = SnakeBidsApp(Path(__file__).resolve().parent.parent)  # to get repository root
+    app = SnakeBidsApp(
+        Path(__file__).resolve().parent.parent,  # to get repository root
+        plugins=[BidsValidator],
```
pvandyken, line 17: The class needs to be instantiated
```suggestion
        plugins=[BidsValidator()],
```

snakebids/plugins/validator.py
```diff
@@ -0,0 +1,58 @@
...
+class InvalidBidsError(SnakebidsPluginError):
+    """Error raised if an input BIDS dataset is invalid."""
+
+
+class BidsValidator:
+    """Perform BIDS validation of dataset, falling back to the pybids
+    version of validation if the node-version of bids-validator is
+    not found.
+
+    Parameters
```
pvandyken, line 21: This docstring belongs under `__call__()`. The main docstring should concern `raise_invalid_bids`

snakebids/plugins/validator.py
```diff
@@ -0,0 +1,58 @@
...
+from snakebids.exceptions import SnakebidsPluginError
+
+_logger = logging.getLogger(__name__)
+
+
+class InvalidBidsError(SnakebidsPluginError):
+    """Error raised if an input BIDS dataset is invalid."""
+
+
+class BidsValidator:
```
pvandyken, line 16: If you're able to, this class could be simplified using attrs

pvandyken (changes requested):
Sorry for another round of comments, I think I just reviewed this too late last night

snakebids/plugins/validator.py
kaitj: I think something like may work?

```
validate=(
    not config.get("plugins.validator.success") and 
    not config.get("plugins.validator.skip")
)
```

| Skip validation _[Config Value]_ | Validator Success _[Config Value]_ | Validate Value |
|-|-|-|
|False _[True]_ | Any _[Any]_ | False |
|True _[False]_ | False _[True]_ | False |
|True _[False]_ | True _[False]_ | True |

pvandyken: The other option is to only link pybids validation to skip. If node validation succeeds, pybids, will succeed for sure. And pybids doesn't strictly validate anyway, it only skips files it doesn't recognize (and that only in the root dataset). So it might even be faster to leave it on.

pvandyken: No real harm in keeping `success`, maybe someone will find it useful. (More likely for detecting validation failures and doing something custom)

There's a possibility enabling validation will make it go faster, because it will retain fewer paths. AFAIK it just looks for a `dataset_description` and checks that paths have valid entitites

kaitj: Latest commit changed the pybids validate logic to only associate with `--skip-bids-validation`. 

I did end up keeping `plugins.validator.success` as you noted. The current use case I can think of is the use of the slightly more complex logic to perform validation with one or the other, but maybe like you said someone else will find it useful for other things.

snakebids/plugins/validator.py
```diff
@@ -0,0 +1,60 @@
...
+class InvalidBidsError(SnakebidsPluginError):
+    """Error raised if an input BIDS dataset is invalid."""
+
+
+@attr.define
+class BidsValidator:
+    raise_invalid_bids: bool = attr.field(default=True)
+
+    def __call__(self, app: SnakeBidsApp) -> None:
+        """Perform BIDS validation of dataset, falling back to the pybids
```
pvandyken, line 23: Looks like this never makes an appearance on the readthedocs reference pages. We should really probably have a new page in addition to 'API' and 'Internals'. If you'd rather just merge this in now, we can make this a new issue.

One thing that might come up though is that I don't know if these dunder methods are currently automatically documented. It might be better to move the bulk of the description to be a class docstring, and keep the `__call__` docstring sparse (sorry that this seems to go against a previous comment... although my previous point about documenting the `app` parameter under `__call__` remains valid). And in this case `raise_invalid_bids` should be documented

snakebids/project_template/{{cookiecutter.app_name}}/workflow/Snakefile
```diff
@@ -13,8 +13,14 @@ inputs = generate_inputs(
     derivatives=config.get("derivatives", None),
     participant_label=config.get("participant_label", None),
     exclude_participant_label=config.get("exclude_participant_label", None),
+    validate=not config.get("plugins.validator.skip", False)
 )
 
+# If skip -> False and __ -> False
+# If validate success -> False and True -> False
```
pvandyken, line 20: Did you mean to keep this table? I don't really understand it either way

pvandyken (approved):
Left one comment RE documentation, but I'd be okay with deferring to a future issue if you'd rather not do it here, I'll leave it up to you

snakebids/project_template/{{cookiecutter.app_name}}/workflow/Snakefile
kaitj: Ahh I did not - that when I was trying to figure out some logic with the previous implementation. Will remove this

snakebids/plugins/validator.py
kaitj: I moved some stuff around and agree about the new page, but I think the new page should probably be introduced in a new PR. Feel free to make additional docstring suggestions.

Once thing I noticed (and something to fix in a separate PR), is the docstring styling is inconsistent in the code base.

Issue #270, opened 2023-04-07
Why is `pybids_db_dir` commented in the boilerplate config?


pvandyken:
I believe use of `--pybidsdb-dir` will set this variable regardless of whether `pybid_db_dir` is commented in the generated config. Since the CLI arg should probably be the preferred use, can we just remove the config entry from the boilerplate app?

akhanf:
I am fine with that

PR #296, opened 2023-05-19
Update pybids_db parameters

### LINKED ISSUE #269: Confusing menagerie of names for pybids database
Currently I think every reference to pybids databases has a different name. We have the parameter in generate_inputs (`pybids_database_dir`), the cli variable (--pybidsdb-dir), and value put into config (`"pybids_db_dir"`). All of these, in turn, are different from the pybids parameter (database_path).

I would propose that `pybidsdb_dir` would be an appropriate name for all of the above (within Snakebids of course). Similar argument for `pybidsdb_reset`.

kaitj:
## Proposed changes

Unifies the parameters across the different modules, config, and CLI. The following have been updated:
* References to the pybids database: `pybidsdb_dir`
* References to resetting the database: `pybidsdb_reset`
_(The one caveat to the reset parameter is that `--reset-db` still exists with a deprecation warning. I'm also open to just deprecating this with this PR.)_

Additionally, this PR removes the commented out lines associated with pybidsdb in the config boilerplate in favor of having the user set the parameters via CLI.

Resolves #269 and #270. 

## Checklist

## Notes


pvandyken (changes requested):
This is great thanks. It probably is still worth having at least one deprecation cycle for the old `generate_inputs` arguments, just to give people time to transition. Don't document them, and print a warning if they're used, and we'll pull them out next release.

kaitj:
More of a note to myself since this was more of a global search and replace:

- [ ] remove mention of in commenting of lines in the documentation.

tkkuehn (approved):
I think a deprecation cycle for `generate_inputs` like Peter said makes sense but have no additional comments, thanks!

snakebids/core/input_generation.py
```diff
@@ -234,14 +240,27 @@ def generate_inputs(  # noqa: PLR0913
...
+    if pybids_database_dir:
+        _logger.warning(
+            "The parameter `pybids_database_dir` in generate_inputs() will be "
+            "deprecated in the next release. To set the pybids database, use the "
+            "`pybidsdb_dir` parameter instead."
+        )
+    if pybids_reset_database:
+        _logger.warning(
+            "The parameter `pybids_reset_database` in generate_inputs() will be "
+            "deprecated in the next release. To resetset the pybids database, use "
```
tkkuehn, line 252: "resetset" -> "reset"

tkkuehn (approved):
One typo to fix but otherwise looks good

snakebids/core/input_generation.py
```diff
@@ -234,14 +240,27 @@ def generate_inputs(  # noqa: PLR0913
         participant_label, exclude_participant_label
     )
 
+    if pybids_database_dir:
+        _logger.warning(
+            "The parameter `pybids_database_dir` in generate_inputs() will be "
+            "deprecated in the next release. To set the pybids database, use the "
```
pvandyken, line 246: This is pedantic, but I believe it's correct to say it's currently deprecated and will be removed in the next release

docs/bids_app/workflow.md
```diff
@@ -15,8 +15,8 @@ configfile: 'config/snakebids.yml'
 inputs = snakebids.generate_inputs(
     bids_dir=config["bids_dir"],
     pybids_inputs=config["pybids_inputs"],
-    pybids_database_dir=config.get("pybids_db_dir"),
-    pybids_reset_database=config.get("pybids_db_reset"),
+    pybidsdb_dir=config.get("pybidsdb_dir"),
+    pybidsdb_reset=config.get("pybidsdb_reset"),
     derivatives=config.get("derivatives"),
```
pvandyken, line 20: Can we move the old versions to the bottom just in case someone was relying on the order? (But we should really make these arguments kw-only in the future).

docs/running_snakebids/overview.md
```diff
@@ -14,8 +14,7 @@ Note that if any rules in the Snakebids workflow use Singularity containers, spe
 
 Indexing of large datasets can be a time-consuming process. Snakebids, through `PyBIDS` has the ability to create or leverage an existing database, requiring indexing of datasets to be only performed when user chooses to do so (usually if the dataset has changed)! Note, this feature is **opt-in**, meaning it is not used unless the associated config variables are used. To opt-in:
```
pvandyken, line 15: The "opt-in" language is maybe a bit confusing? The only sense it's opt-in is that you need to manually connect the config variables to `generate_inputs`. This can probably be cleaned up a bit

pvandyken (changes requested):

docs/running_snakebids/overview.md
kaitj: I tried to reword this part a little bit, but not sure about the "disabled by default". Think we can probably phrase it slightly better.

snakebids/cli.py
```diff
@@ -138,10 +138,18 @@ def create_parser(include_snakemake: bool = False) -> argparse.ArgumentParser:
...
+        "--pybidsdb-reset",
+        "--pybidsdb_reset",
+        action="store_true",
+        help=("Reindex existing PyBIDS SQLite database"),
+    )
+
+    # To be deprecated
     standard_group.add_argument(
         "--reset-db",
         "--reset_db",
```
pvandyken, line 155: Use `help=argparse.SUPPRESS` to hide the old argument from help

snakebids/app.py
```diff
@@ -155,8 +155,8 @@ def run_snakemake(self) -> None:
 
         # Update config with pybids settings
         if args.pybidsdb_dir:
-            self.config["pybids_db_dir"] = args.pybidsdb_dir
-        self.config["pybids_db_reset"] = args.reset_db
+            self.config["pybidsdb_dir"] = args.pybidsdb_dir
```
pvandyken, line 158: I don't think it's a problem if `self.config["pybidsdb_dir"]` is set to `None`? Can we just set it unconditionally?

pvandyken (changes requested):
Just left a few additional comments, should be good to go after

pvandyken (approved):

PR #300, opened 2023-05-23
Add DuplicateComponentError for more precision

### LINKED ISSUE #299: Need to Narrow down the `ValueError` thrown in `BidsDataset.from_iterable()`
When this function is called in `generate_inputs`, it's wrapped in an exception guard that's supposed to detect if multiple components have the same name. But because all of the components are passed in via a lazily-evaluated generator, we end up catching `ValueErrors` from `BIDSLayout`, leading to inaccurate errors. This is easily resolved by using a one-off, purpose-built exception in our own code.

tkkuehn:
## Proposed changes

Resolves #299.

## Checklist

## Notes

### DIFF
--- a/snakebids/core/input_generation.py
+++ b/snakebids/core/input_generation.py
@@ -19,7 +19,7 @@ from snakebids.core.filtering import filter_list
 from snakebids.exceptions import ConfigError, DuplicateComponentError, PybidsError
 from snakebids.types import InputsConfig, ZipList
 from snakebids.utils.snakemake_io import glob_wildcards
-from snakebids.utils.utils import BidsEntity, BidsParseError, MultiSelectDict, surround
+from snakebids.utils.utils import BidsEntity, BidsParseError, MultiSelectDict
 
 _logger = logging.getLogger(__name__)
 
@@ -269,8 +269,7 @@ ses-{session}_run-{run}_T1w.nii.gz",
         dataset = BidsDataset.from_iterable(bids_inputs, layout)
     except DuplicateComponentError as err:
         raise ConfigError(
-            "Multiple components found with the same name: "
-            + ", ".join(surround(err.args[0], "'"))
+            f"Multiple components found with the same name: {err.duplicated_names_str}"
         ) from err
 
     if use_bids_inputs:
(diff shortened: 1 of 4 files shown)


pvandyken (approved):
Thanks Tristan!

snakebids/core/input_generation.py
```diff
@@ -267,7 +267,7 @@ def generate_inputs(  # noqa: PLR0913
 
     try:
         dataset = BidsDataset.from_iterable(bids_inputs, layout)
-    except ValueError as err:
+    except DuplicateComponentError as err:
         raise ConfigError(
             "Multiple components found with the same name: "
             + ", ".join(surround(err.args[0], "'"))
```
pvandyken, line 273: Actually, does this line still play nicely with the `__init__` function you made for the new Exception? The previous `ValueError` just carried an iterable of strings

tkkuehn: Oh shoot, good catch. I'll get this fixed.

snakebids/exceptions.py
```diff
@@ -16,3 +23,15 @@ def __init__(self, msg: str, *args: object) -> None:
...
     """Exception raised when pybids encounters a problem."""
+
+
+class DuplicateComponentError(Exception):
+    """Raised when a dataset is constructed from components with the same name."""
+
+    def __init__(self, duplicated_names: Iterable[str]):
+        self.duplicated_names_str = ", ".join(surround(duplicated_names, "'"))
+        super().__init__(
+            "A BidsDataset cannot be instantiated from BidsComponents with the same "
```
pvandyken, line 34: I think I'd move this nice error message up into the `generate_inputs` function and leave the Exception itself as simple as possible, just because it lives a context-free namespace.

tkkuehn: I don't disagree that this message lives weirdly far from the place where it'll ever get raised, but the counterargument I've seen (from Tryceratops by way of ruff) is that setting up an exception this way gets you a more concise stack trace because you don't also see the error message at the raise site, and that this enforces that the exception always has a consistent message.

Both of these seem like minor pluses to me, but I'm open to just reverting this to match the other exceptions if you'd prefer to maintain the simplicity of `exceptions` here and keep the actual error message close to its actual occurrence.

pvandyken: Sounds good to me

kaitj (approved):

pvandyken (approved):

Issue #287, opened 2023-05-02
`--filter_{input_type}` ValueError


kaitj:
## The problem

Encountered this one this morning while testing a workflow where `--filter-T1w` is used at the command-line. This was previously working in `0.7.2`, but no longer working as of `0.8.0`. Ended up getting the following error:

```python
Traceback (most recent call last):
  File "/project/6050199/tkai/shared_venvs/scattr-X3WkDvET-py3.9/bin/scattr", line 6, in <module>
    sys.exit(main())
  File "/scratch/tkai/0_dev/scattr/scattr/run.py", line 19, in main
    app.run_snakemake()
  File "/home/tkai/project/shared_venvs/scattr-X3WkDvET-py3.9/lib/python3.9/site-packages/snakebids/app.py", line 161, in run_snakemake
    update_config(self.config, args)
  File "/home/tkai/project/shared_venvs/scattr-X3WkDvET-py3.9/lib/python3.9/site-packages/snakebids/app.py", line 257, in update_config
    pybids_inputs[input_type]["filters"].update(arg_filter_dict)
ValueError: too many values to unpack (expected 2)
```

using the following call:

```bash
scattr /scratch/tkai/0_sandbox/scattr_test/bids /scratch/tkai/0_sandbox/scattr_test/derivatives participant --slurm_tmpdir --dwi_dir /scratch/tkai/0_sandbox/scattr_test/derivatives/prepdwi_v0.0.13/ --freesurfer_dir /scratch/tkai/0_sandbox/scattr_test/derivatives/fmriprep_v21.0.0_freesurfer/ --fs-license $FS_LICENSE_FILE --sl-count 5000 -p --cores 8 --use-singularity --bzero_thresh 15 --force-output --verbose --filter-T1w desc=synthstrip
```

I don't think we changed how the `update_config` call worked. May just be how I am calling it, but wanted to raise the issue in case it is something bigger.

## Steps to reproduce

* Run a workflow that uses snakebids `0.8.0` with `--filter-{input_type}` flag.

## Screenshots

If you have any helpful screenshots or a file you tried to upload, please include them!

## Environment

* `snakebids` version (if applicable): `0.8.0`
* Operating system (e.g. Ubuntu 20.04, Windows 10): Ubuntu 20.04

pvandyken:
Probably a good time to handle #94, which is probably the reason this slipped past in the first place.

pvandyken:
[...]

PR #305, opened 2023-06-07
Test magic args

### LINKED ISSUE #94: More tests needed for magic arguments
As discussed in #93, `add_dynamic_args()` and `update_config` are both completely untested. Unit tests should be written.

tkkuehn:
## Proposed changes

Closes #94. The tests aren't super comprehensive but definitely increase the code coverage to the point that they would have caught #287.

### DIFF
--- a/snakebids/tests/strategies.py
+++ b/snakebids/tests/strategies.py
@@ -64,7 +64,16 @@ def input_configs(draw: st.DrawFn) -> InputConfig:
     )
 
     wildcards = draw(st.one_of(st.lists(bids_entity().map(str)), st.none()))
-    custom_path = draw(st.one_of(st.text(), st.none()))
+    custom_path = draw(
+        st.one_of(
+            st.text(
+                alphabet=st.characters(
+                    blacklist_categories=("Cs",), blacklist_characters=("\x00",)
+                )
+            ),
+            st.none(),
+        )
+    )
 
     pybids_inputs: InputConfig = {}
     if wildcards is not None:
--- a/snakebids/tests/strategies.py
+++ b/snakebids/tests/strategies.py
@@ -52,7 +52,7 @@ def bids_entity_lists(
 
 
 @st.composite
-def input_config(draw: st.DrawFn) -> InputConfig:
+def input_configs(draw: st.DrawFn) -> InputConfig:
     filtered_entities = draw(st.one_of(st.lists(bids_entity()), st.none()))
     filters = (
         {
@@ -83,8 +83,8 @@ def input_config(draw: st.DrawFn) -> InputConfig:
     return pybids_inputs
 
 
-def inputs_config() -> st.SearchStrategy[InputsConfig]:
-    return st.dictionaries(st.text(min_size=1), input_config())
+def inputs_configs() -> st.SearchStrategy[InputsConfig]:
+    return st.dictionaries(st.text(min_size=1), input_configs())
 
 
 @st.composite
--- a/snakebids/tests/strategies.py
+++ b/snakebids/tests/strategies.py
@@ -13,7 +13,7 @@ from hypothesis import assume
 from snakebids.core.datasets import BidsComponent
 from snakebids.core.input_generation import BidsDataset
 from snakebids.tests import helpers
-from snakebids.types import ZipList
+from snakebids.types import InputConfig, InputsConfig, ZipList
 from snakebids.utils.utils import BidsEntity, MultiSelectDict
 
 _Ex_co = TypeVar("_Ex_co", bound=str, covariant=True)
@@ -49,6 +49,42 @@ def bids_entity_lists(
         unique=True,
         # bids_paths aren't formed correctly if only datatype is provided
     ).filter(lambda v: v != ["datatype"])
+
+
+@st.composite
+def input_config(draw: st.DrawFn) -> InputConfig:
+    filtered_entities = draw(st.one_of(st.lists(bids_entity()), st.none()))
+    filters = (
+        {
+            entity.entity: draw(
+                st.one_of(st.booleans(), bids_value(), st.lists(bids_value()))
+            )
+            for entity in filtered_entities
+        }
+        if filtered_entities is not None
+        else None
+    )
+    wildcard_entities = draw(st.one_of(st.lists(bids_entity()), st.none()))
+
+    wildcards = (
+        [entity.entity for entity in wildcard_entities]
+        if wildcard_entities is not None
+        else None
+    )
+    custom_path = draw(st.one_of(st.text(), st.none()))
+
+    pybids_inputs: InputConfig = {}
+    if wildcards is not None:
+        pybids_inputs.update({"wildcards": wildcards})
+    if filters is not None:
+        pybids_inputs.update({"filters": filters})
+    if custom_path is not None:
+        pybids_inputs.update({"custom_path": custom_path})
+    return pybids_inputs
+
+
+def inputs_config() -> st.SearchStrategy[InputsConfig]:
+    return st.dictionaries(st.text(min_size=1), input_config())
 
 
 @st.composite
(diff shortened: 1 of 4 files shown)


snakebids/tests/strategies.py
```diff
@@ -51,6 +51,42 @@ def bids_entity_lists(
     ).filter(lambda v: v != ["datatype"])
 
 
+@st.composite
+def input_config(draw: st.DrawFn) -> InputConfig:
```
pvandyken, line 55: It reads a little weird here, but the convention for strategies is to pluralize the name, so I'd change this to `input_configs`

snakebids/tests/strategies.py
```diff
@@ -51,6 +51,42 @@ def bids_entity_lists(
...
+    filtered_entities = draw(st.one_of(st.lists(bids_entity()), st.none()))
+    filters = (
```
pvandyken, line 57: Did you have a reason for not using `st.dictionaries` here? It obviously leads to the same outcome, but the native function may have some optimizations (although I'm not sure).

snakebids/tests/strategies.py
```diff
@@ -51,6 +51,42 @@ def bids_entity_lists(
...
+        {
+            entity.entity: draw(
+                st.one_of(st.booleans(), bids_value(), st.lists(bids_value()))
+            )
+            for entity in filtered_entities
+        }
+        if filtered_entities is not None
+        else None
+    )
+    wildcard_entities = draw(st.one_of(st.lists(bids_entity()), st.none()))
```
pvandyken, line 67:
```suggestion
    wildcard_entities = draw(st.one_of(st.lists(bids_entity()).map(str), st.none()))
```

This will eliminate the need for the below comprehension

snakebids/tests/test_app.py
```diff
@@ -33,6 +38,64 @@ def app(mocker: MockerFixture):
...
+        ),
+    )
+    def test_magic_args(
+        self,
+        filters: tuple[BidsEntity, Iterable[str]] | None,
+        wildcards: Iterable[BidsEntity] | None,
+        path: str | None,
+    ):
+        config_copy = copy.deepcopy(config)
+        config_copy["bids_dir"] = "root"  # type: ignore
```
pvandyken, line 71: If you annotate `config_copy` as `dict[str, Any]`, you can eliminate the type ignores

snakebids/tests/strategies.py
```diff
@@ -51,6 +51,42 @@ def bids_entity_lists(
...
+    if wildcards is not None:
+        pybids_inputs.update({"wildcards": wildcards})
+    if filters is not None:
+        pybids_inputs.update({"filters": filters})
+    if custom_path is not None:
+        pybids_inputs.update({"custom_path": custom_path})
+    return pybids_inputs
+
+
+def inputs_config() -> st.SearchStrategy[InputsConfig]:
```
pvandyken, line 86: As per above, this would be `inputs_configs`

snakebids/tests/test_app.py
```diff
@@ -33,6 +38,64 @@ def app(mocker: MockerFixture):
...
+        )
+        update_config(config_copy, args)
+        inputs_config: InputsConfig = cast(InputsConfig, config_copy["pybids_inputs"])
+        if filters:
+            assert (
+                inputs_config["bold"].get("filters", {}).get(filters[0].entity)
+                == filters[1]
+            )
+        if wildcards:
+            for entity in [wildcard.entity for wildcard in wildcards]:
```
pvandyken, line 93: You can do a set subtraction here:
```suggestion
            assert not {wildcard.entity for wildcard in wildcards} - set(inputs_config["bold"].get("wildcards", []))
```
Alternatively a length comparison on the original vs an intersection

snakebids/tests/test_app.py
```diff
@@ -33,6 +38,64 @@ def app(mocker: MockerFixture):
     return app
 
 
+class TestUpdateConfig:
+    @given(
```
pvandyken, line 42: You should be able to use your `input_configs` strategy here. Just pull whatever filters, wildcards, and path it gives you

snakebids/tests/test_app.py
```diff
@@ -33,6 +38,64 @@ def app(mocker: MockerFixture):
...
+        st.one_of(
```
pvandyken, line 43: This is more stylistic, but I pretty much always prefer explicit keyword arguments for given. It just makes things clearer and can save headache down the road (test debugging, etc)

snakebids/tests/test_cli.py
```diff
@@ -73,6 +77,38 @@ class TestAddDynamicArgs:
     ]
     mock_all_args = mock_basic_args + mock_args_special
 
+    @given(sb_st.inputs_config())
+    @settings(suppress_health_check=[HealthCheck.function_scoped_fixture])
+    def test_dynamic_inputs(self, mocker: MockerFixture, pybids_inputs: InputsConfig):
+        p = create_parser()
```
pvandyken, line 83: This is more of a mental note to myself, haha, but I'm assuming the reason mocker works here without you calling `mocker.stopall()` is because you're not using the `spy` method, just `patch`

snakebids/tests/test_cli.py
```diff
@@ -73,6 +77,38 @@ class TestAddDynamicArgs:
...
+        add_dynamic_args(p, copy.deepcopy(parse_args), pybids_inputs)
+        magic_filters = list(
+            itertools.chain.from_iterable(
```
pvandyken, line 86: Style point: my convention has been to import `itertools as it`, if you want consistency with the rest of the repo

snakebids/tests/test_cli.py
```diff
@@ -73,6 +77,38 @@ class TestAddDynamicArgs:
...
+        )
+        mocker.patch.object(
+            sys,
+            "argv",
+            self.mock_all_args + magic_filters + magic_wildcards + magic_path,
+        )
+        args = parse_snakebids_args(p)
+        for key in pybids_inputs:
+            key_identifier = key.replace("-", "_")  # argparse does this
+            assert isinstance(args.args_dict[f"path_{key_identifier}"], str)
```
pvandyken, line 108: Why do you switch from key access to get?

pvandyken (changes requested):
Thanks for implementing the inputs_config strategies, that'll be useful down the road

snakebids/tests/test_app.py
tkkuehn: Going from here, I landed on using the `<=` set operator

snakebids/tests/test_cli.py
tkkuehn: Ah totally forgot I was supposed to use `mocker.stopall()` generally

snakebids/tests/test_cli.py
tkkuehn: No idea, on reflection. I standardized these on key access.

tkkuehn:
GH actions is being weirdly flaky here but all the tests passed on the last run, so I think this is good for another look

snakebids/tests/test_cli.py
pvandyken: I don't think it's a huge deal overall, it just took me a couple minutes to make sense of this, so I wanted to make some sort of note, haha

snakebids/tests/test_app.py
pvandyken: Ah, cool. Never realized that was an option

pvandyken:
Looks like the failures may have to do with the `custom_paths` strategy generating invalid examples. Check out https://pypi.org/project/hypothesis-fspaths/ for generating only valid paths. (I'm not seriously worried about users using null-bytes in their custom_paths, and if they do, that's on them, so I don't think we need to handle it).

pvandyken:
Ah, sorry, looks like you did take some steps to address the failures. Either way, the library presented is another option (I may incorporate it in the future for other reasons, so we have it on standby at least)

pvandyken (approved):

Issue #205, opened 2022-12-07
Saving more provenance information


pvandyken:
## The problem

This issue is somewhat relate to #182, but more specifically concerns what information is saved in the auto-generated `snakemake.yml` file. Currently we keep the entire config, any arguments passed to `run.py` (including snakemake args), and the snakemake and output directories. For full replicability, it would also be helpful to save the apṗ version (related to #85), and the snakemake version used to run the pipeline.

akhanf:
I agree!

For the app version would we need to force users to expose the version in a particular way?

pvandyken:
One more related thought: so long as we're reconstructing the config file from scratch, and given the order of attributes is arbitrary, it might be nice to purposefully organize things (and even put comments in if that can be managed... not sure) to make the file as human-readable as possible.

akhanf:
yes a more human-readable config would be great!

PR #308, opened 2023-06-15
Save more provenance information

tkkuehn:
## Proposed changes

Addresses #205 in a reasonably basic way. It's not guaranteed that a Snakebids app even has a version, so this PR just takes a crack at it by assuming the name of the snakemake directory is also the name of an installed package. If anyone has any other ideas, we can straightforwardly add fallback options to `SnakeBidsApp.get_app_version`.

Note that this doesn't do anything about the proposed formatting/comments in the output config. Maybe that would be good to raise as a separate issue.

## Checklist

## Notes


kaitj:
Quickly skimmed the review (will do a more thorough look later on), and so far it looks good.

Would there be a situation where the app isn't a distribution package installed in the environment?

tkkuehn:
> Would there be a situation where the app isn't a distribution package installed in the environment?

An example would be if I find a Snakebids app I want to use on GitHub, clone it, and have Snakebids etc. globally installed in my environment, so I don't bother installing the app to my environment.

A real-world (if kind of weird and likely to change) case is https://github.com/afids/afids-CNN/tree/main/afids_cnn, which wraps a couple of different Snakebids workflows under one package that is not itself a Snakebids workflow.

kaitj:
Fair enough, then I think one useful thing to add would be a warning that "...the version of the snakebids app cannot be found as it is likely not installed in the environment" or something to that effect, so its not a surprise that the config shows "Unknown" version.

snakebids/app.py
```diff
@@ -240,6 +250,21 @@ def create_descriptor(self, out_file: PathLike[str] | str) -> None:
         )
         new_descriptor.save(out_file)  # type: ignore
 
+    def get_app_version(self) -> str | None:
```
pvandyken, line 253: I would set up version as a class attribute and turn this function into a factory. That way a user can easily override with their own version if they prefer. And they can easily access the version later via `app.version` (this will help a version plugin later too)

snakebids/app.py
```diff
@@ -240,6 +250,21 @@ def create_descriptor(self, out_file: PathLike[str] | str) -> None:
...
+        """Attempt to get the app version, returning None if we can't.
```
pvandyken, line 254: If we're moving in this direction, I wonder if we should allow (and encourage) the passing of explicit modules (or string names of modules) to `snakemake_dir` in addition to paths. This would actually simplify the `run.py` file a bit. And this would create some extra liberty e.g. if the snakemake package was in some subpackage `foo.bar`, we'd be able to support that a little better.

pvandyken: Not necessarily a goal for this PR, but maybe something to think about for the future

snakebids/tests/test_app.py
```diff
@@ -142,6 +148,9 @@ def test_runs_in_correct_mode(
                 "pybidsdb_reset": True,
                 "snakefile": Path("Snakefile"),
                 "output_dir": outputdir.resolve(),
+                "snakemake_version": metadata.version("snakemake"),
+                "snakebids_version": "0.0.0",  # poetry-dynamic-versioning
```
pvandyken, line 152: `0.0.0` is only the case if you don't have poetry-dynamic-versioning installed as a plugin (e.g. this test fails on my machine). Off the top of my head I can't think of an "independent" way of getting the snakebids versioning without repeating production code, so as you did with `snakemake_version`, you may be stuck calling `metadata.version("snakebids")` here

pyproject.toml
```diff
@@ -52,6 +52,7 @@ boutiques = "^0.5.25"
 more-itertools = ">=8,<10"
 cached-property = "^1.5.2"
 pvandyken-deprecated = "0.0.3"
+importlib_metadata = [ { version = "^6.6.0", python = "<3.8" } ]
```
pvandyken, line 55: In preparation for dropping python37, could you pin this to `==1.4`, which seems to be the version included by default in py38, and make sure everythin works?

pvandyken (changes requested):
One concern with leaving a warning if no version is found is that if a developer messes up the implementation preventing versioning, users will get that warning and potentially think something is wrong... But then they may still think something is wrong when they read `version=unknown` so maybe it doesn't matter.

snakebids/app.py
tkkuehn: I didn't address this for now, but is probably worth a separate issue for discussion.

tkkuehn:
I added a warning where I tried to convey that the missing version will often be nothing for the end user to worry about, but I'm open to changing either or both of the warning's text, and its log level (could make it `INFO`, but those won't usually ever be seen by end users, as discussed in #184 ).

kaitj (approved):
This lgtm - the warning might be a little long, but I think it conveys what needs to be conveyed and I don't have any other suggestions there.

pvandyken (changes requested):
Just one comment RE the warning, but once that's resolved no need for re-review

pvandyken:
@tkkuehn, we had originally said we'd try to get this one in our 0.9 release. Would you still like to? If not, there's no need to maintain py37 support

tkkuehn:
> We had originally said we'd try to get this one in our 0.9 release. Would you still like to? If not, there's no need to maintain py37 support

I'm open either way, I don't think this is such a major feature that we need to get it in ASAP, and delaying reduces the amount of work to be done to drop py37 I guess

pvandyken:
How about we just do a race. If the py37 drop is done before this, I'll release without this, otherwise this can go in

pvandyken:
Nope, looks like just a merge, then you can go ahead

PR #309, opened 2023-06-15
Boolean CLI filters

### LINKED ISSUE #258: Extend boolean filtering to cli
With #191, booleans can be used in the config file to filter for the presence or absence of entities. This capability was never extended into the cli.

There's a bit more ambiguity built in here: does `--filter-suchandsuch space=false` mean filter for spaces called `"false"`, or exclude `space` as an entity? I would lean toward the former and try to find an alternative syntax.

`False` can be expressed with a blank string: `--filter-suchandsuch space=`. `True` is a bit harder, but possibly an asterisk (like a bash wildcard?): `--filter-suchandsuch space='*'`?

tkkuehn:
## Proposed changes

Resolves #258. REQUIRED and NONE just set the `pybids_inputs` to `True` and `False`, respectively, and let Snakebids parse them as usual. OPTIONAL sets the relevant filter to a sentinel value that gets picked up and drops any existing filters.

## Checklist

## Notes

### DIFF
--- a/snakebids/app.py
+++ b/snakebids/app.py
@@ -20,6 +20,7 @@ from snakebids.cli import (
     parse_snakebids_args,
 )
 from snakebids.exceptions import ConfigError, RunError
+from snakebids.types import OptionalFilter
 from snakebids.utils.output import write_config_file  # type: ignore
 from snakebids.utils.output import prepare_bidsapp_output, write_output_mode
 
@@ -253,7 +254,12 @@ def update_config(config: dict[str, Any], snakebids_args: SnakebidsArgs) -> None
     for input_type in pybids_inputs.keys():
         arg_filter_dict = args[f"filter_{input_type}"]
         if arg_filter_dict is not None:
-            pybids_inputs[input_type]["filters"].update(arg_filter_dict)
+            pybids_inputs[input_type].setdefault("filters", {})
+            for entity, filter_ in arg_filter_dict.items():
+                if filter_ == OptionalFilter:
+                    pybids_inputs[input_type]["filters"].pop(entity, None)
+                else:
+                    pybids_inputs[input_type]["filters"][entity] = filter_
         del args[f"filter_{input_type}"]
 
     # add cmdline defined wildcards from the list:
@@ -261,6 +267,7 @@ def update_config(config: dict[str, Any], snakebids_args: SnakebidsArgs) -> None
     for input_type in pybids_inputs.keys():
         wildcards_list = args[f"wildcards_{input_type}"]
         if wildcards_list is not None:
+            pybids_inputs[input_type].setdefault("wildcards", [])
             pybids_inputs[input_type]["wildcards"] += wildcards_list
         del args[f"wildcards_{input_type}"]
 
--- a/snakebids/app.py
+++ b/snakebids/app.py
@@ -256,7 +256,7 @@ def update_config(config: dict[str, Any], snakebids_args: SnakebidsArgs) -> None
         if arg_filter_dict is not None:
             pybids_inputs[input_type].setdefault("filters", {})
             for entity, filter_ in arg_filter_dict.items():
-                if filter_ == OptionalFilter:
+                if filter_ is OptionalFilter:
                     pybids_inputs[input_type]["filters"].pop(entity, None)
                 else:
                     pybids_inputs[input_type]["filters"][entity] = filter_
--- a/snakebids/cli.py
+++ b/snakebids/cli.py
@@ -37,22 +37,21 @@ class FilterParse(argparse.Action):
             return
 
         for pair in values:
-            match_ = re.fullmatch("([a-zA-Z0-9]+).(OPTIONAL|REQUIRED|NONE)", pair)
-            if match_:
-                key = match_.group(1)
-                spec = match_.group(2)
-                if spec == "OPTIONAL":
-                    value = OptionalFilter
-                elif spec == "REQUIRED":
-                    value = True
-                elif spec == "NONE":
-                    value = False
-                else:
-                    # Should never happen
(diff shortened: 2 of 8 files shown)


snakebids/app.py
```diff
@@ -253,14 +254,20 @@ def update_config(config: dict[str, Any], snakebids_args: SnakebidsArgs) -> None
     for input_type in pybids_inputs.keys():
         arg_filter_dict = args[f"filter_{input_type}"]
         if arg_filter_dict is not None:
-            pybids_inputs[input_type]["filters"].update(arg_filter_dict)
+            pybids_inputs[input_type].setdefault("filters", {})
+            for entity, filter_ in arg_filter_dict.items():
+                if filter_ == OptionalFilter:
```
pvandyken, line 259: I think we want `is` here
```suggestion
                if filter_ is OptionalFilter:
```

snakebids/cli.py
```diff
@@ -36,8 +37,25 @@ def __call__(
             return
 
         for pair in values:
-            # split it into key and value
-            key, value = pair.split("=")
+            match_ = re.fullmatch("([a-zA-Z0-9]+).(OPTIONAL|REQUIRED|NONE)", pair)
```
pvandyken, line 40: Unless there's a good reason not to, I would prefer the optional, required, etc be case-insensitive

snakebids/cli.py
```diff
@@ -36,8 +37,25 @@ def __call__(
...
+            if match_:
```
pvandyken, line 41: I think it would simplify this code to just do a split on `.`, as we do for the equal sign below (and it would be a bit faster, for what it's worth). I don't think a valid entity name could ever contain a dot, and either way, this would be equivalent to the code you have currently. Unless there was a reason I'm not aware of for using regex.

pvandyken: Since I can imagine a weird universe where valid entity-values have dots, we would need to check for the equal sign first, then check for a dot.

snakebids/cli.py
pvandyken: Also, following pybids, `ANY` is an alias for `REQUIRED`

snakebids/cli.py
```diff
@@ -214,7 +232,9 @@ def add_dynamic_args(
     # create filter parsers, one for each input_type
     filter_opts = parser.add_argument_group(
         "BIDS FILTERS",
-        "Filters to customize PyBIDS get() as key=value pairs",
+        "Filters to customize PyBIDS get() as key=value pairs, or as "
+        "key.{REQUIRED|OPTIONAL|NONE}, to enforce the presence or absence of values for"
```
pvandyken, line 236: Above I noted that `ANY` should be an alias for `REQUIRED`, but I don't necessarily think it's necessary to list both aliases in the help message here... I leave that to your judgment

snakebids/tests/test_cli.py
```diff
@@ -107,6 +107,59 @@ def test_dynamic_inputs(self, mocker: MockerFixture, pybids_inputs: InputsConfig
             assert isinstance(args.args_dict[f"filter_{key_identifier}"], dict)
             assert isinstance(args.args_dict[f"wildcards_{key_identifier}"], list)
 
+    @given(pybids_inputs=sb_st.inputs_configs())
+    @settings(suppress_health_check=[HealthCheck.function_scoped_fixture])
```
pvandyken, line 111: fyi, just because this settings call is common, I did make a one-liner called `tests.helpers.allow_tmpdir`. Called such because I was using it solely for tmpdirs, but if you want to use it feel free to rename it, as clearly it's useful for other purposes

(Just note that you can't stack `@settings`... so with more complicated settings you need the explicit decorator)

snakebids/tests/test_cli.py
```diff
@@ -107,6 +107,59 @@ def test_dynamic_inputs(self, mocker: MockerFixture, pybids_inputs: InputsConfig
...
+    def test_required_filters(self, mocker: MockerFixture, pybids_inputs: InputsConfig):
+        p = create_parser()
+        add_dynamic_args(p, copy.deepcopy(parse_args), pybids_inputs)
+        magic_filters = list(
+            it.chain.from_iterable(
+                [[f"--filter-{key}", "entity.REQUIRED"] for key in pybids_inputs]
```
pvandyken, line 117: Just an idea for testing case-insensitivity (if this gets implemented), you could try the hypothesis regex strategy to generate variations of the `.suffix` with various cases

snakebids/app.py
```diff
@@ -253,14 +254,20 @@ def update_config(config: dict[str, Any], snakebids_args: SnakebidsArgs) -> None
...
+                else:
+                    pybids_inputs[input_type]["filters"][entity] = filter_
         del args[f"filter_{input_type}"]
 
     # add cmdline defined wildcards from the list:
     # wildcards_{input_type}
     for input_type in pybids_inputs.keys():
         wildcards_list = args[f"wildcards_{input_type}"]
         if wildcards_list is not None:
+            pybids_inputs[input_type].setdefault("wildcards", [])
```
pvandyken, line 270: Does this line ever end up being tested (do our test suites fail if removed?)

pvandyken (changes requested):
One general point I'm still not sure of is the appropriateness of `.` as the seperator character. It's not a character really used for assignment type operations... it implies a hierarchical relationship between the entity and the flag being applied that doesn't really exist. 

The only alternative I have is `:`, which does imply assignment and, purely subjectively, looks better on the CLI. The disadvantage here is potential confusion ("Do I use `=` to assign a flag, or `:`?"). 

Ultimately I suppose it's not a huge issue... users can learn to use whatever we decide. So I won't die on this issue of others have a strong opinion

pvandyken (approved):

PR #314, opened 2023-06-23
Allow Booleans in filter lists

pvandyken:
Previously, we only checked if the entire filter was set to a boolean, not if one of the entries was a boolean. This fix allows us to filter datasets where an entity must either be absent or set to a specific value.


pvandyken:
Ruff was starting to complain about the number of branches in `_get_lists_from_bids`, so I began to refactor some of it into a new class containing steps. I didn't want to spend a ton of time on it, but it's all private API, so we can refine it later as we wish.

pvandyken:
@tkkuehn @kaitj fyi test failures are now resolved

kaitj (approved):
This lgtm, thanks! The refactor due to the Ruff complaint was a nice bonus - a little bit cleaner.

Issue #313, opened 2023-06-23
Use PEP 621-based pyproject.toml in snakebids create


tkkuehn:
## The problem

The `setup.py` we create isn't super useful and is setuptools-specific, so we should use something a bit more standard (and incidentally maybe move toward support for hatch) by implementing a pyproject.toml that has the fields defined in PEP 621.

kaitj:
Think this may have been brought up during the initial discussions too and not opposed to moving towards supporting hatch, but this poetry PR is one we should also keep an eye on as they move towards PEP 621. 

https://github.com/python-poetry/poetry-core/pull/567

pvandyken:
This has been resolved by #188

PR #171, opened 2022-06-02
moved project_template workflow files to package dir

akhanf:
## Proposed changes

- fixes bug in resolving path in run.py
- this location is preferred as it allows for proper inclusion of the
workflow files in any python packaging that is built
- also made run.py executable

## Checklist

- [ ] Changes have been tested to ensure that fix is effective or that a feature works.
- [ ] Changes pass the unit tests
- [ ] I have included necessary documentation or comments (as necessary)
- [ ] Any dependent changes have been merged and published

## Notes
All PRs will undergo the unit testing before being reviewed. You may be requested to explain or make additional changes before the PR is accepted.

PR #316, opened 2023-06-26
Update cookiecutter

### LINKED ISSUE #307: Update app creation instructions
## The problem

After some discussion during the dev lunch, some things that would be helpful:
* Mentioning `pipx` installation of snakebids to be able to use `snakebids create`
* `snakebids create` is mentioned once in the documentation, but may want to make it more visible (maybe part of a quick start guide?)

~I noticed we have a cookiecutter template that is talked about in the documentation, but no actual instructions on how to use it to create a Snakebids project. This may be helpful for new users or those unfamiliar with cookiecutter. At the moment, one would need to call the following:~

~EDIT: Scratch most of this - after talking with Tristan, realized there is a `snakebids create` that is calling the cookiecuuter function. Still wonder about the possibility of splitting it off into its own repo.~

kaitj:
## Proposed changes

Closes #307. Just a small PR to update the cookiecutter template a little bit (will save more major changes related to #313 in another PR) and tried to include a bit more instructions related to creating a snakebids app. Added a few new cookiecutter variables to (e.g. `_bids_version`), but open to changing those or making them a user prompt. I filled in some of the prompts with default values. Also made the documentation generation optional.

This also incorporates some of the changes from the #171 draft.

## Checklist

## Notes


pvandyken:
@kaitj is this good for review or did you want to resolve the test failure first?

kaitj:
Content-wise good for a look. 

It passed 3.8 tests locally, and 3.11 here, so not sure where the issue stems from. I'll track it down when I get a chance!

kaitj:
Ahh I thought I included that. Must have been a different file / repo. Well I’ll change it back but yeah nothing else in this changes.

snakebids/project_template/cookiecutter.json
```diff
@@ -1,10 +1,12 @@
 {
-  "full_name": "",
+  "full_name": "FirstName LastName",
```
pvandyken, line 2: Not sure if this is the case here, but I'd prefer not having sort of meaningless defaults like "FirstName". IMO values should either be mandatory (but is this even possible in cookiecutter??), or optional and completely left out (the way "github" is currently handled).

That being said, I know I personally don't have the mental bandwidth to deal with cookiecutter right now, so if you'd rather just leave it this way in absence of a better solution I'm fine with that.

pvandyken: I'll just leave this as a throw-away for now, but we can perhaps discuss later: I found a "competitor" to cookiecutter called [copier](https://github.com/copier-org/copier) that resolves a lot of the frustrations I have with cookiecutter: good documentation, supports mandatory arguments, uses yaml configuration, less verbose syntax, help messages, validation, etc. It may be worth switching over if it will give our users (and us) a better experience

pvandyken (approved):
Two comments neither of which necessarily need immediate action. If you merge, I'll try to bridge them into seperate PRs

snakebids/project_template/cookiecutter.json
kaitj: > IMO values should either be mandatory (but is this even possible in cookiecutter??), or optional and completely left out (the way "github" is currently handled).

AFAIK, it's not possible to have mandatory arguments and definitely agree about leaving some of the values out.

> I'll just leave this as a throw-away for now, but we can perhaps discuss later: I found a "competitor" to cookiecutter called [copier](https://github.com/copier-org/copier) that resolves a lot of the frustrations I have with cookiecutter: good documentation, supports mandatory arguments, uses yaml configuration, less verbose syntax,

I'm open to switching from cookiecutter. There are definitely a few things I would change if I could having gone through a bit of deeper dive now. Dealing with the variables definitely requires some thought and I'm not completely happy with the way it is.

snakebids/project_template/{{cookiecutter.__app_name}}/docs/getting_started/installation.md
```diff
@@ -0,0 +1,91 @@
...
+run times for each job. Workflows may also contain append images for quality assurance or to
+summarize outputs, by using the `report(...)` function on any snakemake output.
+
+To generate a report, run:
+
+```bash
+{{cookiecutter.__app_name}} /path/to/bids/dir /path/to/output/dir participant --report
+```
+
+## Compute Canada Instructions
```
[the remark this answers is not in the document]
kaitj: Ahh yeah, only ting I did here was switch it from `.rst` to `.md` and move it into the folder, but yeah, we probably want to remove the ComputeCanada instructions for the more general audience. It may be good to go through these template docs more thoroughly, but maybe that would be better saved following a discussion about the competitor you mentioned.

tkkuehn (approved):

PR #322, opened 2023-08-04
Drop support for Python 3.7

pvandyken:
This removes any py37 references and:

- Updates all `__position_only` arguments to `position_only, /`
- Refactors code to use `:=` walrus operators where I found that made sense
- Removes any workaround put in previously

Also updated the dependency specs on `numpy` and `scipy`, and removed for pandas. The idea is to restrain to versions where a wheel is available, since numpy and scipy don't maintain wheels for every active python version. Hopefully it doesn't cause any problems


snakebids/types.py
```diff
@@ -3,17 +3,20 @@
...
 from pathlib import Path
-from typing import Dict, Generic, Iterable, List, Mapping, Sequence, overload
-
-from typing_extensions import (
+from typing import (
     TYPE_CHECKING,
+    Dict,
+    Generic,
+    Iterable,
+    List,
```
kaitj, line 11: Can we update the use of `Dict` and `List` to `dict` and `list` or would that break functionality?

kaitj (approved):
Overall I think looks good to me. There is the testing time out again (this may be a non issue now, but was there when I first started reviewing) and also have 1 relatively minor comment.

snakebids/types.py
pvandyken: We need to keep `Dict` around because it's used for subclassing. `List` can go I think

tkkuehn (changes requested):
Looks good, but needs a merge from main to incorporate the provenance stuff and (maybe?) fix the test failure.

pvandyken:
The debug statements precommit hook was failing on my machine here with py38 syntax (not sure why, it works elsewhere). Rather than fight it, I just dropped it and activated the debugger checks in ruff, which do the same thing.

tkkuehn (approved):

PR #326, opened 2023-08-16
Fix version fetching in readthedocs build

pvandyken:
In stable builds, readthedocs is not currently fetching git tags, which we need for determining project version. So we're getting unhelpful versions at the top of our project documentation like `0.0.1.devxxx-blahblah`.

This is an attempt to fix this by adding some post_checkout steps

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -18,9 +18,18 @@ formats:
   - pdf
 
 build:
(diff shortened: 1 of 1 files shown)


kaitj:
Oh, I definitely misunderstood this when you first mentioned this today. Thought it was happening on the PR builds.

kaitj (approved):

PR #328, opened 2023-08-21
Update docs dependencies

pvandyken:
Updates the documentation dependencies to use sphinx 7. The furo theme is causing problems with the current `requirements.txt`, apparently due to a regression in the way in works with sphinx 6. Rather than go through the effort of raising an issue, it was easier to just update everything. Unfortunately, this does mean we'll need a new 0.9.x release just in order to have a working docs

### DIFF
--- a/.github/workflows/release_drafter.yml
+++ b/.github/workflows/release_drafter.yml
@@ -21,7 +21,8 @@ jobs:
           echo "Comments: ${{ github.event.inputs.comments }}"
 
       - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


tkkuehn (approved):
Looks good to me. Not sure what's going on with the test failure, but I suspect it's just some transient issue with the poetry download

pvandyken:
> Not sure what's going on with the test failure, but I suspect it's just some transient issue with the poetry download

Had to constrain the poetry version to one that still supported py37. I'll try to remember to drop the requirement when porting to `main`

PR #330, opened 2023-08-23
Update dependencies and pyright version

pvandyken:
Fix a number of type issues arising with latest pyright

Need to lock pyright to >=1.1.324 because previous versions have a false
positive bug

Some dict.get(..., {}) require laborious workarounds because the value
type of the dict is a TypedDict, which is different than a regular dict
and can't be constructed easily on the fly. The latest pyright version
uses the latest typeshed, which changed the inference rules to prevent
our previous "hack" from working. This situation will probably never
change.

Broadened the annotated types for some keyword args on functions
accepting both fixed and generic kwargs with differing types. The fixed
kwargs need to allow at least the same types as the generic args, or we
get type issues downstream. Use type validation to compensate

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,7 +70,7 @@ black = "^23.1.0"
 pytest = "^7.0.0"
 pytest-mock = "^3.7.0"
 isort = "^5.10.1"
-poethepoet = "^0.21.1"
+poethepoet = "^0.22.0"
 pre-commit = "^3.0.0"
 mkinit = "^1.0.0"
 hypothesis = "^6.34.1"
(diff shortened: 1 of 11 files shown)


pyproject.toml
```diff
@@ -80,8 +80,8 @@ pyfakefs = "^5.1.0"
 pyparsing = "^3.0.9"
 # Version 1.1.312-1.1.315 have a false positive handling some nested function
 # calls
-pyright = ">=1.1.307,<1.1.312"
-ruff = "^0.0.280"
+pyright = "<=1.1.324"
```
tkkuehn, line 83: I think this sign is the wrong way around.

tkkuehn (approved):
One small fix but this otherwise looks fine to me. Unfortunate that we can't properly type hint our fixed keyword args anymore, I guess!

kaitj (approved):

PR #331, opened 2023-08-23
Add poetry lock check

pvandyken:
The poetry lock file is quite frequently becoming inconsistent with the
pyproject.toml, so add a PR check to ensure they remain synchronized.

Also update the python quality runner to 3.11, and add some fixes to
make the test workflow work with the ACT gh-actions simulator


tkkuehn (commented):
Maybe a naive question, but where does this add the lock check?

pvandyken:
Oh no, haha! Thanks for catching, I had accidentally reset my git state before pushing and forgot to actually put the check back in

tkkuehn (approved):

PR #332, opened 2023-08-23
Refactor filter_list to use RegexContainer

pvandyken:
The original intent of the PR was to add boolean support to `filter_list` so
that we could support boolean filters with custom paths, but I relealized after
some initial refactoring that such support doesn't really make any sense. So
now the PR has just the refactors:

This simplifies the filtering operation and will allow for more flexible
control of regex vs regular filtering in the future

RegexContainer is a container containing a regex expression. Strings are
"contained" if the "re.match" the expression

Move ContainerBag from test code to utils for use in filter_list

### DIFF
--- a/snakebids/utils/utils.py
+++ b/snakebids/utils/utils.py
@@ -10,7 +10,9 @@ from os import PathLike
 from pathlib import Path
 from typing import (
     Any,
+    AnyStr,
     Callable,
+    Container,
     Generic,
     Iterable,
     Iterator,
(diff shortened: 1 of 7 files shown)


snakebids/utils/utils.py
```diff
@@ -539,3 +514,36 @@ def index(
...
+    def __contains__(self, x: object, /):
+        if isinstance(x, type(self.template.pattern)):
+            return self.template.match(x) is not None
+        return False
+
+
+class ContainerBag(Container[_T]):
+    """Container to hold other containers
+
+    Useful because list(Container) isn't guarenteed to work, so this lets us merge
```
tkkuehn, line 538: "guarenteed" -> "guaranteed"

tkkuehn (approved):
One very minor typo otherwise this looks good

kaitj (approved):

PR #334, opened 2023-08-24
Add deprecation mechanism to old pybidsdb vals 

### LINKED ISSUE #333: `pybids_db_dir` was eliminated without warning, causing migration issues
In #296, we consolidated all the database related terms into one single name: `pybidsdb_dir`. We added a helpful deprecation warning to the old `pybids_database_path` parameter in `generate_inputs`, however, we failed to implement any deprecation cycle for the config entry: `pybids_db_dir`. If you run a workflow with snakebids>=0.9, that entry will not be set.

In all of my workflows, I had a line in my `generate_inputs` call:

```
pybids_database_dir=config.get("pybids_db_dir")
```

After upgrading, snakebids would silently fail to read my database (since it was just getting `None` to `pybids_database_dir`). This has stymied me a couple of times already, and I was one of the people behind the rename!

I think it would be really good to have some sort of deprecation, but unfortunately it will be a bit tricky. In the pattern above, we really can't tell if the user is selecting `pybids_db_dir` out of the config dict. The only way I thought would be to set that variable to some magic value which we can detect. It feels brittle, but at least it would be a temporary measure until we fully deprecate the parameter. And it's arguably better than the current situation.

Any thoughts on this or any better ideas? @tkkuehn @kaitj. If I make a change I'll push it onto the 0.9.x branch so that people have a chance of finding it.

pvandyken:
The resolved path was getting overwritten by the args_dict in
SnakebidsArgs. This didn't happen previously because the resolved path
was saved under a different name than the CLI parameter, but since
resolving the names, the CLI parameter name overwrites the resolved
name.

Fix by changing the order of update_config and the manual setting of the
pybids path

Resolves #333

### DIFF
--- a/snakebids/utils/utils.py
+++ b/snakebids/utils/utils.py
@@ -50,6 +50,10 @@ class BidsTag(TypedDict):
 BidsTags: TypeAlias = "dict[str, BidsTag]"
 
 
+DEPRECATION_FLAG = "<!DEPRECATED!>"
+"""Sentinel string to mark deprecated config features"""
+
+
 @ft.lru_cache(None)
 def read_bids_tags(bids_json: Path | None = None) -> BidsTags:
     """Read the bids tags we are aware of from a JSON file.
--- a/snakebids/app.py
+++ b/snakebids/app.py
@@ -28,7 +28,7 @@ from snakebids.exceptions import ConfigError, RunError
 from snakebids.types import OptionalFilter
 from snakebids.utils.output import write_config_file  # type: ignore
 from snakebids.utils.output import prepare_bidsapp_output, write_output_mode
-from snakebids.utils.utils import to_resolved_path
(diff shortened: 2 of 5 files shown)


kaitj (approved):
This largely lgtm

snakebids/utils/utils.py
```diff
@@ -50,6 +50,10 @@ class BidsTag(TypedDict):
 BidsTags: TypeAlias = "dict[str, BidsTag]"
 
 
+DEPRECATION_FLAG = "<!DEPRECATED!>"
+"""Sentinal string to mark deprecated config features"""
```
tkkuehn, line 54: "Sentinal" -> "Sentinel"

tkkuehn (approved):
One typo in a docstring but otherwise looks good to me

PR #336, opened 2023-09-15
Overhaul `snakebids create`

pvandyken:
NB: I pushed PR a bit early because I want to test the gh-action changes. Pending bug fixes there the only thing missing is an additional test on the `pyproject.toml` writing

Use copier instead of cookiecutter:

As per previous discussion, but in summary, it has
- Better feature set: in addition to everything cookiecutter has, it supports mandatory values, custom validation, more Q/A types, post-completion messages, etc
- Better documentation
- More polished (e.g. typed codebase)

Change starting workflow into a "welcome" workflow:
- Instead of the previous fMRI workflow, we now have a simple workflow with just the `all` rule. When run, the rule will print out a welcome message saying everything works and giving a link to the tutorial
- Within the `tests` folder, the generated app will have a set of empty image files matching the tutorial, so users can use this `create`d app directly while following the tutorial (tutorial updates saved for future PR)

Instead of the previous `setup.py` file, which made no one happy, the setup form gives choice of build system. The choices are `setuptools`, `poetry`, `hatch`, and `flit` (I'm unaware of any other major build systems). Of those choices, all except `poetry` work with the [PEP621](https://peps.python.org/pep-0621/) `pyproject.toml`, so the only section we need to customize is `[build-system]`. `poetry` gets its own `pyproject.toml`. `setuptools` additionally requires a `MANIFEST.in` to ensure all the non-python files get included in the build package.

Build systems are separate from environment managers, and there's simply too many possible ways to install dependencies (e.g. with/without `pyenv`, which manager tool, etc), so that's left to the user. To assist, however, `create` will print out some suggested instructions depending on the build-tool chosen. For `poetry`, we give `poetry` specific instructions. For all the others, we give instructions for `setuptools`, `hatch`, and `pdm`. `flit` can do dependency management, but not environment management, so I don't give any instructions for it.

Finally, I've implemented end-to-end testing of app creation, dependency installation, and initial
run. To preserve the environment of any machine performing these tests, I've set them up to run in docker containers. Unfortunately, this does add some technical burden to get the tests to work, but it seemed simpler than all the alternatives. Tests requiring the docker container automatically skip if docker or the container is not present. The container will not be published anywhere, but should be built when needed. For local machines, I have a `poe` command for this, for GH-actions, I've setup the action to build and cache.

When reviewing, please clone the branch and give the command and follow-up a try! The main point here is to make the user experience as smooth as possible.


pvandyken:
Should be ready for review @tkkuehn @kaitj

snakebids/admin.py
```diff
@@ -57,7 +84,7 @@ def gen_parser() -> argparse.ArgumentParser:
 
 
 def main() -> None:
-    """Invoke Cookiecutter on the Snakebids project template."""
+    """Invoke snakebids cli."""
```
kaitj, line 87: Minor nitpick, but just for consistency: snakebids -> Snakebids

snakebids/project_template/copier.yaml
```diff
@@ -0,0 +1,140 @@
...
+create_doc_template:
+  type: bool
+  help: Would you to set up basic documentation?
+
+bids_version:
+  default: "1.8.0"
+  when: false
+
+snakebids_version:
+  default: "0.9.2"
```
kaitj, line 69: Think this was brought up during the dev meeting, but just wanted to double check about the version updating here and how the dynamic versioning / version bumping affects this default.

snakebids/project_template/{% if build_system != 'poetry' %}pyproject.toml{% endif %}.jinja
```diff
@@ -0,0 +1,48 @@
...
+        {%- endif -%}
+    {{ ' ' }}}
+]
+{%- endif %}
+classifiers = [
+    "Programming Language :: Python :: 3",
+    "Operating System :: OS Independent",
+]
+
+requires-python = ">=3.8,<3.12"
```
kaitj, line 29: Just a minor point - can this be set as a private variable and plugged in here? From a maintenance perspective, this is easier to update the version in 2 places than in all the different jinja templates if we end up supporting more Python versions in the future (assuming we always keep this up-to-date with what is supported in Snakebids). 

From a user perspective, one possibility is to ask the User about this, but I don't necessarily think this is needed since this is easy enough to update in the generated file afterwards.

kaitj (approved):
Thanks for making this overhaul and updating the tests! LGTM with a few nitpicky comments that don't necessarily need to be addressed.

pyproject.toml
```diff
@@ -80,8 +82,10 @@ pyparsing = "^3.0.9"
 # Version 1.1.312-1.1.315 have a false positive handling some nested function
 # calls
 pathvalidate = "^3.0.0"
-pyright = ">=1.1.324"
+# As of pyright==1.1.327 only 1.1.324 is bug free
```
tkkuehn, line 85: Is there a specific issue we need to see fixed to get rid of this pinned version? Would be good to make a note of that

snakebids/tests/conftest.py
```diff
@@ -66,4 +67,5 @@ def bids_fs(fakefs: FakeFilesystem | None) -> FakeFilesystem | None:
         fakefs.add_real_file(f / "bids.json")
         fakefs.add_real_file(f / "derivatives.json")
         fakefs.add_real_file(Path(*resources.__path__) / "bids_tags.json")
+    print(Path(*hypothesis.vendor.__path__))
```
tkkuehn, line 70: This feels like it was for debugging and got left in by mistake. If not, what is it?

snakebids/tests/helpers.py
```diff
@@ -261,7 +263,27 @@ def expand_zip_list(
...
-    return dict(zip(it.chain(zip_list.keys(), new_values.keys()), zip(*new_cols)))
+    z = zip(*new_cols)
+    return dict(zip(it.chain(zip_list.keys(), new_values.keys()), z))
+
+
+def needs_docker(container: str):
+    def inner(func: _F) -> _F:
+        try:
+            sp.run(["docker"], check=True)
+        except sp.CalledProcessError:
```
tkkuehn, line 274: I think this will actually raise a (python builtin) `FileNotFoundError`

pyproject.toml
pvandyken: Once `pyright==1.1.328` release we can upgrade to that one. The main code is out, just waiting for the `pip` package to catch up (and VSCode). Has to do with the way it analyzes nested `zips`, of which we have many

snakebids/tests/conftest.py
pvandyken: Ah, quite right, thanks

snakebids/tests/helpers.py
pvandyken: True, I'll listen for both then. On `WSL`, the docker command only works if the Windows side `docker-desktop` is running, so we have a situation where the `docker` command can be found but won't actually do anything.

snakebids/project_template/{% if build_system != 'poetry' %}pyproject.toml{% endif %}.jinja
pvandyken: Yep!

snakebids/project_template/copier.yaml
pvandyken: Good catch, this needs to be fixed. I think I'll add a variable to the environment via jinja extension that returns the current snakebids version. That should hopefully make it work both on installed and non-installed setups

snakebids/project_template/copier.yaml
```diff
@@ -0,0 +1,126 @@
...
+  help: Provide a brief description of your app
+
+app_version:
+  default: "0.1.0"
+  help: Starting version number for your app
+  validator: '{% if not app_version %}Required{% endif %}'
+
+build_system:
+  help: >
+    What build system would you like to use? If you don't know what this means,
```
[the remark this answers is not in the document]
pvandyken: I can do 2. For 1., already `setuptool` is the first option, meaning the selector defaults to it (so if you just press enter, you get setuptools). Is that what you mean?

pvandyken:
@tkkuehn @kaitj question: one thing I'm uncomfortable with is that I'm skipping the docker tests when docker is not around so that devs without a working docker setup can still run the test suite. However, on CI we don't want the docker tests to accidentally skip. And even on a dev machine, for WSL because you need Docker Destop on windows running, it's easy to accidentally be missing docker when you thought you had it. In that case, I'd rather have a test failure than a silent skip. Any idea how to work around this (maybe a mark `ensure-docker` or something?)

snakebids/project_template/copier.yaml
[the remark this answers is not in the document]
kaitj: Ahh yeah, I guess I either missed that part or it wasn't clear when when you had demo'd this last week, but good to know it already is the selector default.

pvandyken:
> Hmm, I agree - having a test failure would be better than a silent skip. I think my preference would be to opt into skipping docker tests (and maybe this is what you had in mind already) - that is:
>  1. Having some sort of mark as you had suggested
>  2. If docker is not installed or unavailable for some reason, have it fail with messaging that indicates how to skip docker tests (e.g. pytest -m not docker)?

Cool, I'm find with this

tkkuehn:
> I think my preference would be to opt into skipping docker tests (and maybe this is what you had in mind already) - that is:
> 
>     1. Having some sort of mark as you had suggested
> 
>     2. If docker is not installed or unavailable for some reason, have it fail with messaging that indicates how to skip docker tests (e.g. `pytest -m not docker`)?

Sounds good to me too!

pvandyken:
@kaitj, @tkkuehn, I've made a few additions since the first push, so I'm requesting a re-review. You can just look at the commits from "Address Comments..." onwards.

Docker tests now run by default, but have a mark allowing their disabling. Failures because of missing dockers leave a helpful error.

I'm autocalculating the snakebids version of the template based on:
1. The latest version we get from pypi api
2. If fails: the version of our installed snakebids dependency
3. If fails, or yields an invalid version (some dev version, or 0.0.0): fall back to a hard-coded version in the `copier.yaml`. I didn't set this to auto-update with `poetry-dynamic-version` because the plan is to split this code into its own repository, making auto-updates irrelevant

kaitj (approved):
Caught spelling in one of the template comments but good otherwise.

snakebids/tests/test_generate_inputs.py
```diff
@@ -556,7 +556,7 @@ def get_subset(of: Iterable[T]) -> list[T]:
 
 class TestCustomPaths:
     @pytest.fixture()
-    def temp_dir(self, fakefs_tmpdir: Path):
+    def temp_dir(self, fakefs_tmpdir: Path, bids_fs: Path):
```
tkkuehn, line 559: What does the added `bids_fs` parameter here do?

tkkuehn (approved):
Just one point of confusion but this looks good to me

snakebids/tests/test_generate_inputs.py
pvandyken: I need to clean this up a bit, basically there's a bunch of real fs paths that need to be added to the fake filesystem, but I have the fakefs setup and path addition split into separate fixtures. The above has been broken for a long time, and the only mystery for me is why the tests ever worked... apparently inter-test environment isolation is not water-tight, because this was revealed I think when I started splitting this tests with `pytest-split`

PR #340, opened 2023-09-21
Remove dependency upper-caps and update

pvandyken:
Remove cached-property dep since functionality is provided by functools in py38

Remove all upper version limits in main dependency block. Where kept, justification comment is added

Upgraded pvandyken-deprecated dependency, which packages proper typing support (thus, remove our own typings file)


kaitj (approved):
Largely lgtm code-wise (just realized the one test is failing for py3.11), with a very minor comment (that upon second thought doesn't really require any action given these are more dev dependencies than for distribution).

Aside from `cookiecutter` (which will get replaced in #336 anyway), is there opposition to changing the `^` to a `>=`?. The caret requirements would be placing an upper-cap on the dependencies if I am interpreting the poetry docs correctly. Making the change would also bring us closer to following [PEP440](https://peps.python.org/pep-0440/).

pvandyken:
> Aside from cookiecutter (which will get replaced in https://github.com/akhanf/snakebids/pull/336 anyway), is there opposition to changing the ^ to a >=?. The caret requirements would be placing an upper-cap on the dependencies if I am interpreting the poetry docs correctly. Making the change would also bring us closer to following [PEP440](https://peps.python.org/pep-0440/).

For a specific dependency, or for everything? And are you asking about removing the `^` from a syntax perspective (and replacing with `>=x,<y`), or just on removing upper limits?

pvandyken:
The test failure should be fixed by a commit I made in #336, so I'll wait for that one to merge, then deal with the failure here

kaitj:
> > Aside from cookiecutter (which will get replaced in #336 anyway), is there opposition to changing the ^ to a >=?. The caret requirements would be placing an upper-cap on the dependencies if I am interpreting the poetry docs correctly. Making the change would also bring us closer to following [PEP440](https://peps.python.org/pep-0440/).
> 
> For a specific dependency, or for everything? And are you asking about removing the `^` from a syntax perspective (and replacing with `>=x,<y`), or just on removing upper limits?

More a general thing and yeah, replacing `^` with `>=x` (and `<y` where it is needed)

tkkuehn (approved):
Pending tests passing this looks good

pvandyken:
I'm going to keep the `^` syntax for the dev dependencies, just because it's more concise and imo upper limits are at least not harmful for dev dependencies. Pending the test passing I'm going to merge

PR #341, opened 2023-09-23
Refactor install steps into resusable workflow

pvandyken:
Completing the Actions refactoring suggested by @tkkuehn.

@kaitj, you're welcome to take the flow and incorporate it into your centralized repository, if you wish

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,42 +18,11 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Clone repo
-        uses: actions/checkout@v3
-
-      # install & configure poetry
-      - name: Cache Poetry Installation
-        id: cached-poetry
-        uses: actions/cache@v3
-        with:
-          path: ~/.local
-          key: ${{ runner.os }}-poetry-${{ github.event.pull_request.id || github.event.after }}
-      - name: Install Poetry
-        uses: snok/install-poetry@v1
-        if: steps.cached-poetry.outputs.cache-hit != 'true'
-        with:
-          virtualenvs-create: true
-          virtualenvs-in-project: true
-
-      - name: Setup Python
-        id: setup-python
-        uses: actions/setup-python@v4
+      - name: Install
+        uses: khanlab/actions/.github/actions/action-setup_task-installPyProject@v0.3.2
         with:
           python-version: '3.11'
-          cache: 'poetry'
-      - name: Store Pip Cache
-        uses: actions/cache@v3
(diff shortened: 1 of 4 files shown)


kaitj (commented):
This largely looks good to me with one question. On pushing new changes to an open PR, does it reference `github.event.pull_request.id` or `github.event.after`? If the latter, would this mean the setup is rerun everytime there is a new push? (Assuming this references the commit hash.)

_~And yes, I will end up throwing it in the centralized repo!~ I've added a modified version of it in the centralized repo, pending a test before merging into main. Had to change it slightly as (I think) it had to be defined as a job rather than a composite action. In any case, I'll ping you over there once I make the PR._

.github/workflows/test.yml
```diff
@@ -74,41 +43,11 @@ jobs:
...
-      with:
-        virtualenvs-create: true
-        virtualenvs-in-project: true
-
-    - name: Set up Python ${{ matrix.python-version }}
-      id: setup-python
-      uses: actions/setup-python@v4
+    - name: Clone repo
+      uses: actions/checkout@v3
+    - uses: ./.github/actions/install-project
```
kaitj, line 48: Just want to double check to see if this exists - I didn't come across this when looking both here at the parent repo or on your forked.

.github/workflows/test.yml
```diff
@@ -18,42 +18,11 @@ jobs:
...
-        if: steps.cached-poetry.outputs.cache-hit != 'true'
-        with:
-          virtualenvs-create: true
-          virtualenvs-in-project: true
-
-      - name: Setup Python
-        id: setup-python
-        uses: actions/setup-python@v4
+      - name: Install
+        uses: khanlab/actions/.github/actions/action-setup_task-installPyProject
```
kaitj, line 22: I can't recall at the moment - wondering if you know what the behaviour is if you don't tag a specific version? I assume it would either use the latest tag or whatever is on `main`?

kaitj (commented):
This largely looks good to me, but there is the one action path that I am not sure if it exists. Once that is confirmed to exist / updated, then this one is good to go.

.github/workflows/test.yml
pvandyken: Nope, just missed this one!

.github/workflows/test.yml
pvandyken: Good call, I'll add a tag

kaitj (approved):
This one is good to go once the `installPyProject` tag is updated.

PR #344, opened 2023-10-18
Overhaul global filtering of participants

### LINKED ISSUE #216: `participant_label` applies subject filter to every single component
When supplying either `participant_label` or `exclude_participant_label`, snakebids creates a new filter that then gets applied indiscriminately to every component, whether or not appropriate. In my case, I discovered this when trying to grab the `participants.tsv` file. This should be possible based on my tests with pybids, but fails because the subjects filter gets applied.

This would also be a problem if the user supplied their own `subjects` filter. Because the snakebids generated `participant_label` filters get applied second, they would override.

I would propose we apply the `participant_label` filter only if the user lists `subject` as a wildcard. If a filter is also provided, we take the intersection of the user specified and `participant_label` lists.

If only a subject filter is provided, only one subject can be specified, otherwise a multiple paths error will be thrown by snakebids. It seems to me that if a user very specifically filters for one subject in their config, they must have a good reason for doing so and that should override whatever is indicated by `participant_label`

Similarly, if `subject` is not specified as either a wildcard or filter, but is nonetheless an entity, there still by definition can be only one subject, otherwise a 'multiple paths' error is thrown. Again, targeting that very specific subject would seem to override whatever concerns `participant_label` may be addressing.

Of course, we could also insist that `participant_label` always be satisfied, even in the above two cases. For the latter, we'd have to quietly check every single component for the subject entity to ensure alignment. I'd be interested in thoughts on this. @tkkuehn, @akhanf, @kaitj 

## Environment

Likely a problem with every snakebids until now

### LINKED ISSUE #303: exclude_participant_label and filter_component incompatible in some cases
## The problem

`exclude_participant_label` turns regex matching on, which is probably not expected when you're filtering a component from the CLI

pvandyken:
--participant-filter and --exclude-participant filter have had a couple of bugs over the past few versions:

1. Both terms would apply to every single component, even components without the subject entity. These components would have all of their entries filtered out
2. Using --exclude-participant-filter would turn on regex matching mode for every single filter. This changed the meaning of the filters (e.g. allowing partial matches), leading to workflow disruption

The overhaul fixes both bugs, while improving the organization of the
input generation code.

Additionally, the documentation states the magic filter `regex_search: True` can be used to enable regex searching for a block of filters. This hasn't worked for the past few versions. Rather than fix it, the behaviour has been silently disabled in preparation for an overhaul of the regex filtering api

Resolves #303
Resolves #216

### DIFF
--- a/docs/bids_app/config.md
+++ b/docs/bids_app/config.md
@@ -9,11 +9,11 @@ Config Variables
 
 ### `pybids_inputs`
 
-A dictionary that describes each type of input you want to grab from an input BIDS dataset. Snakebids will parse your dataset with {func}`generate_inputs() <snakebids.generate_inputs>`, converting each input type into a {class}`BidsComponent <snakebids.BidsComponent>`. The value of each item should be a dictionary with keys ``filters`` and ``wildcards``.
+A dictionary that describes each type of input you want to grab from an input BIDS dataset. Snakebids will parse your dataset with {func}`generate_inputs() <snakebids.generate_inputs>`, converting each input type into a {class}`BidsComponent <snakebids.BidsComponent>`. The value of each item should be a dictionary with keys `filters` and `wildcards`.
 
-The value of ``filters`` should be a dictionary where each key corresponds to a BIDS entity, and the value specifies which values of that entity should be grabbed. The dictionary for each input is sent to the [PyBIDS' get() function ](#bids.layout.BIDSLayout). `filters` can be set according to a few different formats:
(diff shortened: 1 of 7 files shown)


kaitj (commented):
This largely looked good to me ~- I just wanted to take another look through it before approving it to make sure I didn't miss anything~. Thanks for tackling this!

kaitj (approved):

pvandyken:
Just want a second review on this one. @Bradley-Karat, @myousif9, are one of you able to take a look?

Bradley-Karat:
Everything looks good to me!

Bradley-Karat (approved):

PR #349, opened 2023-12-11
Transition to versioned bids functions

pvandyken:
## TODOS

- [x] test bidspec mocking docstring_formatter as unavailable to ensure docstring still added
- [x] exclude `if TYPE_CHECKING:` from codecov

Another large PR! The task ended up being more finicky than expected, but I'll try to justify some of my decisions below:

## Version functions using configuration, rather than import

The versioned imports (e.g. `import bids_v0_0_0 as bids`) would have worked well in our snakemake context, where we typically directly import the `bids` function. However, it doesn't scale well to any other context. There's three particular problems to point out:

1. If `bids()` is only imported once, as in snakemake, all is well, but in a standard python project, where bids must be repeatedly imported, now you have to ensure you import the same version in every file
1. I'm hoping one day to have `bids`-like methods directly on `BidsComponent` that will derive new paths based on the old component. The versioned imports wouldn't help configure these methods. 
1. The long-term goal is to move path building methods from `snakebids` into `rsbids`. `rsbids` will have a much more robust configuration system, and the versioned imports wouldn't fit well in it.

So I decided instead to make the bids spec globally configurable via `snakebids.set_bids_spec`. The function takes a spec object (importable via `snakebids.paths.specs.vx_x_x()`), or the string name of one of the official specs (e.g. "vx_x_x"). Obviously, this introduces non-pure, state-dependent behaviour for `bids()`, but I think this type of thing is very typical for configuration in python libraries (e.g. numpy, matplotlib, etc all use similar global config objects).

## Dynamic generation of python code and interface files

In an effort to reduce the impressive levels of code duplication, especially docstrings, that would have been required for the versioned bids functions, I used a lot of dynamic importing in the path module. Basically, you can define `__getattr__` directly in a module, and that will be used for dynamic imports. Then, to avoid reading a bunch of specs that would never get used, I transitioned the entire project to lazy imports (made possible by a custom fork of `mkinit` that is awaiting a [PR](https://github.com/Erotemic/mkinit/pull/40)). Finally, to keep type checking working, I set up a script to automatically generate `*.pyi` files for the dynamically generated versioned functions.

And then... I decided to switch over to the configurable system. So now the above paragraph only pertains to the spec reading functions (in `snakebids.paths.specs`), which are still dynamically generated and versioned. So I'm keeping around all the code generation stuff, but if it seems like overkill, that's why.

## Documentation updates

I did some reorganization of the documentation, converting the very flat structure into a more explicit "User guide" and "API" sections. The main api page has been subdivided into different subpages based on category. Unfortunately, the API is complex enough that I could no longer sanely rely on autogen to make the pages, so most of the main functions are explicitly list in the documentation source (although the "Internal" API page is mostly automatic). So something to keep in mind for future maintenance

## Blanket deprecation of `include_subject_dir` and `include_session_dir`

My original idea was to keep these arguments only for the `v0_0_0` spec and not have them for all future specs. But the logic to maintain that exception was getting too complex, so I decided to still keep those arguments for every spec, but deprecate them, with plans for removal in 1.0. I don't know that many people use these arguments anyway (maybe in some of @akhanf's workflows?)


akhanf:
Nice work! Haven't gone through what you've done yet, but looks really cool at a glance!

re: the deprecation of `include_subject_dir` and `include_session_dir` that sounds fine, I use it rarely enough that I am good to either keep those uses pinned to an early snakebids version, or adapt the code moving forward.. By the way, what is the preferred way to use e.g. the subject entity in a filename only and not in the folder?

pvandyken:
At this point a custom spec:

```
bids_factory(specs.v0_0_0(sub_dir=False))(subject="001")
```

If you need such paths frequently, otherwise use `Path(...).name`. 

Is that sufficient? We can consider a more privileged API if it's sufficiently useful, but I'm hoping the above methods are enough.

pvandyken:
One other thing to discuss before this is ready, I still want the bids function to return a `Path` object rather than a string (if/when this gets ported over to `rsbids`, it would return a `BidsPath`). I know I tried this before and broke some workflows (probably because there's workflows doing `+` operations?). So one, do people think this is a worthwhile goal to begin with, and two, what would be an acceptable transition strategy? If we go with this, the two likely options would be 1) keep a permanent exception for the v0_0_0 spec, and 2) keep a `str` return until `1.0`, and then hard switch everything. 1) is more permanent, 2) is much easier to implement

kaitj:
Wow - a lot of great work done here! Will wait for the "ready for review" before going through it. 

To your discussion / questions:

1. I personally agree with returning a `Path` object, based on my own use. That said, I think this might be a discussion that is worth calling a dev meeting for to get some more opinions of others who are working with Snakebids. 
2. I think if we are make the switch, trying to keep a permanent exception might be a softer transition. The hard switch I think may cause workflows to break unexpectedly when the Snakebids upgrade ultimately done as you noted from your previous attempts.

pvandyken:
> I think if we are make the switch, trying to keep a permanent exception might be a softer transition. The hard switch I think may cause workflows to break unexpectedly when the Snakebids upgrade ultimately done as you noted from your previous attempts.

Yeah, so with a hard switch, it definitely wouldn't be until 1.0. Basically any workflows that break with the change would have to pin their snakebids version until they can be fixed

kaitj:
Hmm, yeah, I guess I wasn't thinking about the type hinting too much there. The more I think about it, I think the hard switch is fine as long as we give ample warning (i.e. have a warning message now that this is going to change come 1.0).

I agree about the redundancy - I'd rather choose one path and go down that way.

snakebids/paths/resources/spec.0.10.1.yaml
```diff
@@ -0,0 +1,78 @@
+version: "v0.10.1"
+description: |
+  Spec corresponding to `BIDS v1.9.0`_.
```
kaitj, line 3: Blanking atm, but is there a need for the `_`?

kaitj (commented):
This largely looks good to me, but given its a fairly large PR, I'll take a second look through to make sure I didn't miss anything before approving.

snakebids/paths/resources/spec.0.10.1.yaml
pvandyken: Yes, `rst` reference syntax. (see link at bottom of description)

snakebids/paths/resources/spec.0.10.1.yaml
```diff
@@ -0,0 +1,78 @@
+version: "v0.10.1"
```
kaitj, line 1: I don't recall if this discussion was had a while back, but just in terms of naming & versioning the specs, I can see the argument both for and against versioning with the `snakebids` version vs `BIDS` spec. The note in the docs is a nice warning (if using the `snakebids` version) that things can change at any time. However, I find that the specs page itself is a bit hidden if I am looking to understand the specs or to compare differences between versions. I think if we are leveraging the `snakebids` version, I'd suggest to have (some versionof) the specs doc page more visible and also to think around how to highlight changes between different spec versions (hard to picture right now given just `v0_0_0` and `v0_10_1`).

The more I think about it, the more I like the way this is versioned, just worried about visibility.

(Sorry, got a little side-tracked from my initial thought and ended up on the docs).

docs/api/paths.rst
```diff
@@ -0,0 +1,52 @@
...
+.. autofunction:: set_bids_spec
+
+.. _specs:
+
+Specs
+-----
+
+.. py:currentmodule:: snakebids.paths.specs
+
+Official specs represent the evolution of the BIDS spec over time. Currently, :func:`~snakebids.bids` uses the :func:`v0_0_0` by default, however, in a future release, this will change to always use the latest spec. Specs may be updated at any time, without warning, even on patch releases, so it's important for production code to explicitly specify the BIDS spec version in use. This is done using :func:`~snakebids.set_bids_spec`:
```
kaitj, line 19: The "Official specs represent the evolution of the BIDS spec over time" is a little bit confusing to me. I am having a bit of difficulty disentangling the snakebids spec vs the BIDS spec. By "official", do you mean the Snakebids specs?

docs/tutorial/tutorial.md
```diff
@@ -247,13 +245,14 @@ After these changes, the workflow should still run just like the last step, but
 
 ```
 
-# Part II: Snakebids
+
+(snakebids_tutorial)=
+## Part II: Snakebids
```
kaitj, line 250: Given the snakebids specs, should we include some mention / usage of it in the tutorial?

kaitj (commented):
Still looks good to me - had a few remaining questions / comments around the specs, but I think once those are cleared up this is good to go (on my end). 

Sorry, one my thoughts got a little derailed (you'll see in the comments), but hopefully it still somewhat makes sense! :sweat_smile:

docs/tutorial/tutorial.md
pvandyken: True, good point. The tutorial will probably evolve quite a bit following #362, so I'm leaning to just omitting, especially since it's not strictly necessary unless you use custom entities, and then I think we have a warning.

snakebids/paths/resources/spec.0.10.1.yaml
pvandyken: Yeah, I think the difficulty with the official spec is that we don't actually follow it, we have a lot extra entities based on what pybids can parse.

Currently the specs description is semi hidden in the API, but there is a reference and link at the bottom of the "Bids Function" page. I'm not so sure how to put it, because it's really mostly important to people using custom entities (since I think it's extremely unlikely we would remove an entity from a future spec). So a dedicated page is too much real estate I think (especially since it's the only "configurable" value). I'm definitely open to another specific suggestion though.

For diff, not sure, ideally it would be automated (the current docs are). Currently you can see all of the entities included in the spec in their order. You could write narrative documentation for each spec summarizing updates from the previous one, release notes style. But let me know if you had something else in mind.

docs/api/paths.rst
pvandyken: I'm seeing here now some strong language about using specs compared to my comments on the PR 😉. I agree this warning seems too hidden for its tone and urgency.

I do in principle still think production code should always specify, but its more of an insurance policy. Maybe I'll change the language to qualify that this is mostly relevant to code using custom entities.

pvandyken: To actually address your point, I'll rework the sentence.

By custom spec you mean user specified? The idea is if you use a custom entity and you want to explicitly position it, you can provide your own spec. But maybe you had something else in mind?

kaitj: Yep, use-specified is what I meant here.

snakebids/paths/resources/spec.0.10.1.yaml
kaitj: I don't really have a good solution for either of the last 2 points. 

For the specs description, I think having it semi hidden in the API should be okay, but I think somewhere in one of the bids (and eventually tutorial?) pages we should probably have it highlighted. You're right that having its own dedicated page might be too much realestate.

For the diff, I think the release notes style may take up too much real estate on its own. Is it possible to do some sort of hidden (dropdown) block with notes? Anyways, just something to think about towards the future updates - might be good to bring up in a dev meeting too to gather more thoughts.

pvandyken: > For the diff, I think the release notes style may take up too much real estate on its own. Is it possible to do some sort of hidden (dropdown) block with notes? Anyways, just something to think about towards the future updates - might be good to bring up in a dev meeting too to gather more thoughts.

Yeah, I'm fine with further discussion. If it's alright, I don't think getting this correct now is crucial to the feature, so I won't let this point stop me from merging

pvandyken:
@kaitj, I expanded on the discussion of bids specs in the documentation. It's definitely not perfect: I don't discuss custom specs, and it's really too much prose now to be in the API section. I'd like to leave addressing of these imperfections for another time though, as mentioned above, the docs will likely be evolving quite a bit moving forward anyway.

kaitj (approved):

PR #350, opened 2023-12-12
Fix formatting of empty zip lists

### LINKED ISSUE #339: `_format_zip_lists` breaks when formatting an empty zip list
As per the title...

pvandyken:
Empty zip lists (meaning zip lists with entities but without any values) gave an `IndexError` when attempting formatting. This fixes the problem, plus refactors the formatting code. In particular, the function to find the elision has been refactored to work only on the 'values' portion of the table to reduce the amount of 'magic numbers', previously needed to account for the table keys

Resolves #339

### DIFF
--- a/snakebids/io/printing.py
+++ b/snakebids/io/printing.py
@@ -5,6 +5,7 @@ import json
 import operator as op
 import textwrap
 from math import ceil, floor, inf
+from typing import Sequence
 
 import more_itertools as itx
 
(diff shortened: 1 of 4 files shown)


kaitj (approved):
Took me a little bit to wrap my head around https://github.com/akhanf/snakebids/blob/b8f78980481904263832ab9e2fa0eedb7e9d3829/snakebids/io/printing.py#L27-L28 and the subsequent code, but looks good to me!

PR #351, opened 2023-12-13
Dep updates and expansion of ruff CI

pvandyken:
Since ruff now supports black-compatible formatting, I decided to drop our `black` and `isort` dependencies and switch fully over to ruff. I also enabled all the ruff rule sets that I thought made sense for this repo. Notably, nothing to do with type checking (pyright handles that). Our code base was already quite conventional, so enabling most of these wasn't too burdensome, mostly a good chance to catch up on docstrings!

I left the ruleset additions sprawled across the commits to make it easier to see what rules enforced what changes. There's no intentional behavioural changes, so assuming the tests pass, I'll just wait for one review here.


snakebids/tests/strategies.py
```diff
@@ -537,32 +537,43 @@ def multiselect_dicts(
...
+        )
 
 
 def _is_hashable(item: Any, /):
     try:
         hash(item)
-        return True
     except TypeError:
         return False
+    else:
```
kaitj, line 566: Does this need the `else:` here or would something like the code block work?

```python
def _is_hashable(item: Any, /):
    try:
        hash(item)
    except TypeError:
        return False
    return True
```

Also similar question for `supports_eq()`

snakebids/tests/test_app.py
```diff
@@ -171,19 +171,20 @@ def test_runs_in_correct_mode(
         )
 
         # Prepare expected config
+        db_path = "/path/to/db"
```
kaitj, line 174: Just want to make sure this generic `db_path` won't cause any issues for the testing (or write somewhere unexpected if testing local)

kaitj (commented):
Thanks for tackling this, also nice to see more support from ruff. Largely looks good to me with a few comments and questions to specific changes.

snakebids/tests/test_app.py
pvandyken: All the io functions are mocked, so nothing gets written. I'll make this clear in a comment

kaitj (approved):

PR #352, opened 2023-12-13
Allow builtins and resolved paths as cli types

### LINKED ISSUE #294: Allow builtins as types
We currently lookup strings passed to the `type` attribute in the yaml CLI specification in the `globals()` object. The primary purpose for this was to enable the setting of CLI params as `Path` so that they could be resolved properly to the output directory. We are currently unable to use builtins such as `int` as `type` because these classes are not found in the `globals()` object.

I propose an addition to the existing logic: if the type cannot be found in `globals` we will additionally search for it within the `builtins` module.

pvandyken:
The architecture for type conversion in the argparse integration was intended for using pathlib.Path, but had no flexibility even for builtin types like int

Here, remove the old `globals()` lookup, which exposed too much internal api. Instead, attempt to lookup simple strings as builtins, and strings with dots as resolved module imports (e.g. `module.submodule.type`). `Path` is kept as an exception, so using `type: Path` will still convert the argument into a path

Resolves #294


snakebids/tests/test_cli.py
```diff
@@ -198,17 +199,71 @@ def test_converts_type_path_into_pathlike(
...
+        with pytest.raises(ConfigError, match="cannot be used as a type"):
+            add_dynamic_args(parser, new_args, {})
+
+    def test_using_class_method_as_type_gives_error(
+        self, parser: ArgumentParser, mocker: MockerFixture
+    ):
+        new_args = {
+            "--new-param": {
+                "help": "Generic Help Message",
+                "type": "snakebids.utils.utils.BidsEnity.from_tag",
```
kaitj, line 261: Should this be `BidsEntity` or was this a mispelt? I am assuming this test would be passing as is due to the caught `ImportError`

docs/bids_app/config.md
```diff
@@ -81,7 +81,26 @@ A mapping from the name of each ``analysis_level`` to the list of rules or files
 
 ### `parse_args`
 
-A dictionary of command-line parameters to make available as part of the BIDS app. Each item of the mapping is passed to [argparse's add_argument function](#argparse.ArgumentParser.add_argument). A number of default entries are present in a new snakebids project's config file that structure the BIDS app's CLI, but additional command-line arguments can be added as necessary.
+A dictionary of command-line parameters to make available as part of the BIDS app. Each item of the mapping is passed to [argparse's `add_argument` function](#argparse.ArgumentParser.add_argument). A number of default entries are present in a new snakebids project's config file that structure the BIDS app's CLI, but additional command-line arguments can be added as necessary.
+
+As in [`ArgumentParser.add_argument()`](#argparse.ArgumentParser.add_argument), `type` may be used to convert the argument to the specified type. It may be set to any builtin type, or the fully resolved path of a derived type. [`Path`](#pathlib.Path) may also be used directly:
```
kaitj, line 86: Could you clarify what you mean by "fully resolved path of a derived type"?

kaitj (commented):
Almost looks good, just had one clarification question and potentially a misspelling in the tests.

snakebids/tests/test_cli.py
pvandyken: Woops, yes, thanks!

pvandyken:
An essential problem with the current fix: with snakemake, every CLI arg will have to be serialized into YAML. Even with paths, we don't actually keep them as paths, they get turned into strings. So I'm going to add checks to ensure that all types selected can be simply serializable into primitive yaml types. I'll keep the module path logic though, as I already have the tests, and it may be useful in a more generic future if snakebids works with tools beyond snakemake

pyproject.toml
```diff
@@ -68,6 +67,7 @@ copier = ">=8.2.0"
 jinja2-time = ">=0.2.0"
 # minimum 2.31.0 because of security vulnerability
 requests = ">=2.31.0"
+ruamel-yaml = ">=0.17.2"
```
pvandyken, line 70: Switched to ruamel because pyyaml didn't support allowing only specified types to be serialized, it either tried to serialize everything (using weird custom python tags that snakemake would then refuse to read), or only serialize basic types (but we need it to specifically support Path). ruamel did support this.

If we switch to reading yaml with ruamel in the future, we could further take advantage of its round-trip comment preservation

snakebids/cli.py
```diff
@@ -204,6 +205,26 @@ def create_parser(include_snakemake: bool = False) -> argparse.ArgumentParser:
     return parser
 
 
+def _find_type(name: str, *, yamlsafe: bool = True) -> type[Any]:
```
pvandyken, line 208: `yamlsafe` is unused, but leaving this more as a hint to the future that we may want to support more flexible parsing of types for different applications

pvandyken:
I left a few specific comments above.

I'll write a test for the config writing function (although this functionality wasn't actually tested before), but should be good for review nonetheless

kaitj (commented):
This one lgtm pending resolving that test failure.

PR #355, opened 2023-12-16
Nightly Testing Suite

pvandyken:
Add a nightly testing action to be run once per week. This will include running all the tests, plus, it checks whether the number of tests has changed since last calculation of runtimes, and if so, recalculates the runtimes for `pytest-split` and makes a PR.

~As always with gh-actions, getting this right will involve some dart throwing, so I'll leave this in draft until everything seems to work.~

Hopefully this works. Its hard to fully replicate the behaviour of it running automatically on the main branch, but we can repair any further issues as they come I think

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,78 @@
+name: Weekly testing and recalibration of test timings
+
+on:
+  schedule:
+    - cron: "5 3 * * 2"
+
+jobs:
+  build-cache-env:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
+    steps:
+      - uses: khanlab/actions/.github/actions/action-setup_task-installPyProject@v0.3.3
+        id: setup
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: setup docker
+        uses: ./.github/actions/build-test-container
+        with:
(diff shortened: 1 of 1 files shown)


pvandyken:
@kaitj, as with the actions refactoring, feel free to use this in the centralized repo, although you'll probably want some automated tests (just to make the workload of generalizing this harder!)

kaitj (approved):
This looks good to me workflow wise. Yeah, testing is always a bit iffy with actions I find, but nothing that can't be fixed! 

Re: Reusable action - I've been thinking around this since I started doing something similar with the testing / codecov with Afids. I've been a bit hesitant to add testing into the centralized repo primarily due to a lack of standardized testing practices across the different repos. Would love to chat about this some more though and discuss ideas. If moving forward, we settle on a common practice, then I definitely agree that these should be included.

pvandyken:
> Re: Reusable action - I've been thinking around this since I started doing something similar with the testing / codecov with Afids. I've been a bit hesitant to add testing into the centralized repo primarily due to a lack of standardized testing practices across the different repos. Would love to chat about this some more though and discuss ideas. If moving forward, we settle on a common practice, then I definitely agree that these should be included.

For sure. Actually, the main thing I was referring to was the code to check if the test timings need to be updated, and run if necessary. That's probably a fairly useful and generic chunk of code for any other repos using pytest-split. But yeah, testing is contextual, so even for that workflow, you'd probably need to make the actual test command one of the inputs

PR #356, opened 2023-12-16
Add test coverage

pvandyken:
As per the title!

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -76,8 +76,17 @@ jobs:
           HYPOTHESIS_PROFILE: pr
         run: >-
           poetry run pytest -n auto --splits 5 --group ${{ matrix.split }}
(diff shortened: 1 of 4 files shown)


kaitj (commented):
This one largely looks good to me - I am wondering if there is a benefit to setting a minimum target vs having it set to `auto`. Something we can revisit in the future too.

pvandyken:
> This one largely looks good to me - I am wondering if there is a benefit to setting a minimum target vs having it set to auto. Something we can revisit in the future too.

I know the current coverage is above 90%, which I'm fine with. Auto will make sure we don't go below that (it just ensures coverage never decreases). A limit would let the coverage decrease so long as it never goes below the limit. So if maintaining test coverage will be too much of a burden with future PRs, a threshold might be better, but I would say snakebids is at a maturity point where every new feature should be tested.

kaitj (approved):

PR #357, opened 2023-12-18
Add tests of pypi installation in nightly workflow

pvandyken:
Rather than just run the usual test workflow with the pinned dependencies, I thought it would be more useful to test installing the project from pypi with the latest of all dependencies. So this contains the attempt to do that. The code depends on a custom branch in khanlab/actions, but if everything works here, I'll make a merge of that branch

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -3,26 +3,18 @@ name: Weekly testing and recalibration of test timings
 on:
   schedule:
     - cron: "5 3 * * 2"
+  push:
+    branches:
+      - '*'
+      - '!push-action/*'
+  pull_request:
+    branches:
+      - '*'
+      - '!push-action/*'
 
 jobs:
-  build-cache-env:
-    runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
-    steps:
-      - uses: khanlab/actions/.github/actions/action-setup_task-installPyProject@v0.3.3
-        id: setup
-        with:
-          python-version: ${{ matrix.python-version }}
-      - name: setup docker
-        uses: ./.github/actions/build-test-container
-        with:
-          python-version: ${{ steps.setup.outputs.python-version }}
-
   calibrate-timings:
     runs-on: ubuntu-latest
-    needs: ["build-cache-env"]
     permissions:
       pull-requests: write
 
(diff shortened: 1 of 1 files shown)


kaitj:
More of a thought before you dive too far into this, wondering about the installing from pypi. How does that change if dependencies change or is there a dev version getting pushed to pypi as well?

pvandyken:
> More of a thought before you dive too far into this, wondering about the installing from pypi. How does that change if dependencies change or is there a dev version getting pushed to pypi as well?

So I'm changing the goal post a bit: I'm going to install the main branch, but use pip instead of poetry (just use poetry for testing deps) so that pinned versions don't get used. This should greatly simplify the workflow

pvandyken:
@kaitj, hopefully I've fixed the permission issue causing the weekly failure on the live version of this workflow

.github/workflows/nightly.yml
```diff
@@ -76,3 +61,59 @@ jobs:
...
+        with:
+          python-version: ${{ steps.setup.outputs.python-version }}
+          load: true
+
+      - name: install lib
+        run: poetry run pip install .
+
+  test:
+    runs-on: ubuntu-latest
+    needs: [ 'build-cache-env'  ]
```
kaitj, line 90: Just being nitpicky here, but there is an extra space between `'` and `]`.

.github/workflows/nightly.yml
```diff
@@ -76,3 +61,59 @@ jobs:
...
+          install-deps-only: dev
+          install-library: false
+
+      - name: build docker test container
+        uses: ./.github/actions/build-test-container
+        with:
+          python-version: ${{ steps.setup.outputs.python-version }}
+          load: true
+
+      - name: install lib
```
kaitj, line 85: More out of curiousity, is there a reason for installing the library after building the docker test container / is this used at all downstream? (I assume the docker test container is making use of that dev-only environment somehow).

.github/workflows/nightly.yml
```diff
@@ -76,3 +61,59 @@ jobs:
...
+      fail-fast: false
+    steps:
+      - uses: khanlab/actions/.github/actions/action-setup_task-installPyProject@v0.3.4
+        id: setup
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache-key: nightly
+          install-deps-only: dev
+          install-library: false
+      - name: install lib
```
kaitj, line 104: Nitpicking to add a line between `install-library: false` and `- name: ...`

kaitj (commented):
Largely looks good to me, just a couple of nitpicky comments and a question.

.github/workflows/nightly.yml
pvandyken: The idea is to cache everything installed by pip. I don't think it matters what order the docker and installation are done

kaitj: Hmm, I'll see if I can dig into this a little bit more - from a brief comparison between the `test` part of the workflow and the the `build-cache-env`, I'm not sure I am seeing where things are getting cached (for the repeated installation steps).

pvandyken: Everything installed by pip gets cached. So the test step inherits the pip cache and will install libs from their where applicable. But not sure if I'm understanding the confusion

kaitj: Ahh, realized I was actually comparing the wrong thing (the workflow checks vs the changes to the nightly workflow). Nonetheless, I am not sure I seeing how building the docker container and installing the lib is caching for the downstream steps. 

In the workflow checks at the moment (without having dug too deep into the logs):

1. The parts of each step using `installPyProject` is making use of the cache. 
2. The docker container gets built, where it restores from a relatively small cache before going through all of the steps of building the container (this is repeated for each test split), where it appears to be downloading the libraries.
3. I don't actually see where the install lib is happening in the logs, but I think I need to dig deeper here.

This might also be a case of workflow checks not using the most up-to-date workflow changes.

pvandyken: The docker container may not be cached properly, but just to be clear, that is a separate issue from this PR. In fact, the nightly workflow is not being run as one of the checks (it's only triggered by the schedule)

kaitj: Agreed that the docker cache is a separate issue, though the pip install (the one where it is being called with install lib), doesn't appear to be cached either (from what I can make out in the logs). We can address these in a separate issue / PR given that this is does add the testing of the PyPI installation and that part at least seems to be working.

kaitj (approved):

.github/workflows/nightly.yml
pvandyken: > though the pip install (the one where it is being called with install lib), doesn't appear to be cached either (from what I can make out in the logs). 

This is what I'm confused about, I'm not running that workflow. (or are you looking at the current weekly run? But I don't think that uses the install lib thing).

pvandyken: I looked at some historic runs, and I do think you were in fact correct. It doesn't look like the cache is being used. But I think this is probably more a shortcoming of the khanlab actions installer... it might not be caching the correct directory.

Docker caching is annoying, but not something I feel super motivated to fix 😒

kaitj: Yeah, it's something we can keep an eye on and fix if needed / eventually.

As far as Docker caching goes, I wouldn't worry about it too much - it doesn't look like it takes too long to build the container. I'll leave this link here as a reference if we do end up trying to look into docker caching:

https://dev.to/dtinth/caching-docker-builds-in-github-actions-which-approach-is-the-fastest-a-research-18ei

PR #358, opened 2023-12-20
Move example dataset from docs into template

pvandyken:
This should solve the problem of the template not being shipped with snakebids

Unfortunately, there's no easy way to test this this works. If you install the app using a project manager as editable, the old symlink wouldn't cause a problem. But when packaging into a wheel, the symlink isn't followed, so the files aren't included. My nightly idea for installing from pypi would've picked up on this, but that ends up being quite tricky to implement so I don't think it's worth it. Hopefully this doesn't come up again!

### DIFF
--- a/docs/tutorial/bids
+++ b/docs/tutorial/bids
@@ -1,0 +1,1 @@
+../../snakebids/project_template/tests/data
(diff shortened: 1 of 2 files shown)


kaitj (approved):
Thanks for looking into this!

PR #359, opened 2023-12-20
Manage doc deps with poetry

pvandyken:
I've been looking forward to this one for a while!

Now that `flake8` (and it's annoying upper pin on docutils) has been removed, we can actually manage sphinx and related dependencies write in our pyproject. So I've added a new docs group and add all the documentation related deps there. Readthedocs has been set to use poetry instead of installing with pip.

As a bonus, we now get a new poe task `poe docs` which uses `sphinx-autobuild` to build the docs and refresh on file changes.


.readthedocs.yml
```diff
@@ -26,9 +26,8 @@ build:
...
     pre_install:
       - git update-index --assume-unchanged docs/conf.py
-
-python:
-  install:
-    - requirements: docs/requirements.txt
-    - method: pip
-      path: .
+    post_create_environment:
+      - pip install poetry
```
pvandyken, line 30: Needs an extra line to install dynamic versioning, otherwise the docs don't display with the right version

kaitj (commented):
This is great - codewise looks good to me. Somewhat minor (as it is for more our previews), I'm not sure if readthedocs rebuilt the preview after your last commit or it handles versioning differently? When I open up the preview, it still shows `0.0.0` (and maybe it was already doing this previously and I forgot).

image

pvandyken:
> This is great - codewise looks good to me. Somewhat minor (as it is for more our previews), I'm not sure if readthedocs rebuilt the preview after your last commit or it handles versioning differently? When I open up the preview, it still shows 0.0.0 (and maybe it was already doing this previously and I forgot).

Yep, sorry this was the reason for my comment, but I made it on my phone so I wasn't super clear

kaitj (approved):

PR #361, opened 2023-12-21
Add testing for Python 3.12

pvandyken:
Also caps snakemake below 8.0 until we can build compatibility (major breaking api changes in that release)

A small fix in the comparison logic in `ImmutableList`


kaitj (approved):

kaitj:
Code-wise this looks good to me, I'm not sure what is happening with the test there failing for one of the splits though.

pvandyken:
> Code-wise this looks good to me, I'm not sure what is happening with the test there failing for one of the splits though.

I have to restrict snakemake to <8.0 in the code (at least I'm pretty sure that's the problem...)

pvandyken:
Also, never added 3.12 to the test matrix!

kaitj:
> Also, never added 3.12 to the test matrix!

Ahh, thought you did - saw the 3.12 in the strategy matrix there. I assumed the check wasn't testing 3.12 because it was pulling the workflow from `main`, but I probably should have checked that.

pvandyken:
Had to drop testing of `pdm` for the template, as it's unable resolve more than two versions of dependencies at the same time. This particularly effects numpy, as there's no numpy version that can be resolved for both python 3.8 and python 3.12. Progress is tracked in [this issue](https://github.com/pdm-project/pdm/issues/46)

snakebids/project_template/copier.yaml
```diff
@@ -65,7 +65,7 @@ bids_version:
   when: false
 
 python_version:
-  default: ">=3.8,<3.12"
+  default: ">=3.8,<3.13"
```
pvandyken, line 68: This needs to be set to the range as currently set on pypi. Next release we'll need to remember to update this to `>=3.8,<4.0`

PR #362, opened 2024-01-10
Major update to bidsapp

pvandyken:
~Full description to come, for now just want the tests to run.~

~Also, won't be addressing any further fixes until a few upstream PRs are merged, as I don't want to do more than one painful merge~

See below


pvandyken:
Okay, I think this is ready to go.

A full explanation of the details shall be reserved for the docs, but in summary, this PR creates the `bidsapp` module with its `bidsapp.app` entrypoint function. The returned `_Runner` class a group of action methods that run various portions of the app lifecyle. This cycle is a very minimal skeleton; the life comes from an extensive network of plugins that should be loaded into the app.

The cycle is as follows:

1) load the config with initial values (e.g. the snakemake plugin uses this to read the config files)
2) add cli arguments to the argument parser (most plugins make use of this). Following this step, CLI arguments will be parsed
3) handle unrecognized arguments (the snakemake plugin uses this to handle snakemake args)
4) handle recognized arguments (giving the opportunity to manipulate the `Namespace` returned from the argument parser). Following this step, all remaining items in the `Namespace` are transferred into config
5) Perform arbitrary actions on the config (the snakemake plugin uses this to set up the output directory)
6) Perform a run (the snakemake plugin uses this to actually run snakemake)

Each of these steps represents potential sites for plugin hook-in. Plugins are managed using `pluggy` from pytest, a mature framework that's simple to use and extend.

The `run.py` template is also changed: instead of an entry function, `bidsapp.app()` is directly assigned to an `app` variable. The advantage here is that this `app` variable could be used in the future as a snakebids specific entrypoint, allowing other apps to discover the workflow regardless of file structure (for instance, bidsapp orchestrator could get direct access to the snakebids api in compatible apps without needing to know details of app internals and without needing to go through the cli). Rest assured `bidsapp.app()` is a very cheap function call (it just scaffolds the runner without actually parsing arguments or anything), so there's no concern about running code when importing. A similar convention is used for typer CLIs.

The `_Runner` class will probably be renamed to something public in the future, once its API is stabilized. For now its actions are only related to running the app, but in the future it could have "dry_run", "doc_gen", or other actions. These would create standardized workflow actions completely independent of the underlying architecture (e.g. snakemake, nipype, etc).

snakebids/plugins/pybidsdb.py
```diff
@@ -0,0 +1,122 @@
...
+        self.try_add_argument(
+            group,
+            "--pybidsdb-reset",
+            "--pybidsdb_reset",
+            action="store_true",
+            dest="reset",
+            help="Reindex existing PyBIDS SQLite database",
+        )
+
+        # To be deprecated
```
kaitj, line 88: Given the large PR, I wouldn't be opposed to deprecating the "to-be-deprecated" pybidsdb args here, but can also see an argument for deprecating in a separate PR.

kaitj (approved):
This looks great and thanks for tackling this! Looking forward to testing it out once this is merged in and released. 

Took me a little bit to wrap my head around everything (even though we had already discussed this over dev lunch). My only comment is related to the to-be-deprecated pybids_db args (see comment).

_Edit: I just realized I was never assigned to this but somehow got a notification for this PR. Wonder how that happened. Also noticed auto-assign didn't work here. I think this PR will probably benefit from a second (or even third) reviewer given the size of it._

snakebids/plugins/pybidsdb.py
pvandyken: Ah, it doesn't assign when I finish a draft

kaitj: We probably don't have it configured properly to do so. I'll take a look into it tomorrow since we have something working in another repo.

pvandyken: Also, to address the original point, the comment is a bit misleading: they are already deprecated, just not removed. But a warning is raised if they are used, and they're not included in help text. I'd like to not eliminate them yet just so that this isn't a breaking release

kaitj: Fair enough - In my mind, without having tested it, I was already treating this already as a breaking release as I would of thought any app using `snakebids<11.x` would not be able to migrate to `11.x+` without changes.

In that case, we can leave them in.

pvandyken: Nope, this is (should be...) completely backward compatible. `snakebids.app` is left in place, just re-implemented using the bidsapp module

pvandyken: Okay, what if I left in the dynamic args as a no-op, and I added a `parser` property to the `SnakeBidsApp` that returned the compiled parser. I think that should work without code modification?

pvandyken: > One thing to note (and I'm not sure if this is just due to me building locally, out-of-date sphinx, sphinx backend, or some other reason) is that when building the docs, it doesn't like the use of Path(__file__).resolve().parent. I had to provide the path as a string (e.g. ../scattr) for it to build. On the flipside, the app was unable to find the config file if I provided the string path.

Is this with the bidsapp update, or has that always been the case?

pvandyken: I checked out your scattr docs, the reason `__file__` doesn't work is because you're using the `filename` attribute in the `argparse` directive. Rather than importing the file, this just calls `exec()` on the contents of the file, so no `__file__` variable gets defined. The fix is to use `ref` as follows:

````md
```{argparse}
---
ref: scattr.run.get_parser
prog: scattr
---
```
````

Comment (kaitj) [reply snakebids/plugins/pybidsdb.py:88]: Ahh fair enough, that did do the trick. I think all of the repos I noted earlier are doing the same thing since this was copied from one of our other repos (more mentioning this so we can refer back to this later on). 

The `run.py` does still require a minor change (as you had previously pointed out) to be able to parse the app options for readthedocs since the dynamic_args is no longer an option. We can always note this in the release notes (or documentation).

Comment (pvandyken) [reply snakebids/plugins/pybidsdb.py:88]: Yeah, so I'm thinking of leaving that dynamic args function around but not doing anything and leaving a warning. I never thought about it because the template never even had that function for the doc generation before. In fact, there's numerous problems with the docs template, I have a new issue #391

Comment (pvandyken) [reply snakebids/plugins/pybidsdb.py:88]: @kaitj I added back the dynamic cli args function. Could you check that this makes for a more graceful change for dependent apps (e.g. scattr)?

Comment (kaitj) [reply snakebids/plugins/pybidsdb.py:88]: Has `app.config` also changed? I've started with a fresh branch to try to drop in the new snakebids changes.

```python 
reading sources... [ 72%] usage/cli                                                                     
Exception occurred:
  File "../scattr/run.py", line 11, in get_parser
    add_dynamic_args(app.parser, app.config["parse_args"], app.config["pybids_inputs"])
TypeError: 'NoneType' object is not subscriptable
```

pvandyken: Mm, yes, I'll update it to retrieve the config from the underlying bidsapp

PR #364, opened 2024-01-12
Restrict pulp version in template app

pvandyken:
The latest versions of pulp dropped a deprecated api snakemake dependended on. Newer versions of snakemake will address this, but older versions don't have an upper pin on pulp.

We cannot lower-pin the snakemake version, as we need older snakemakes to maintain support for python < 3.11.

This is not an ideal situation, as if we want the template app to work on all python versions supported by snakebids, and if we want to support all actively maintained pythons, we won't be able to drop this pin on pulp until python 3.10 is dropped, which won't be for a few years.

Probably the best is to wait until either 3.8 or 3.9 is EOL, then change the template app to require python >= 3.11 and snakemake >= 8.2. By that point, python 3.13 or 3.14 will be released, so people should have been able to get their default python version up to 3.11 or higher.

### DIFF
--- a/snakebids/project_template/{% if build_system != 'poetry' %}pyproject.toml{% endif %}.jinja
+++ b/snakebids/project_template/{% if build_system != 'poetry' %}pyproject.toml{% endif %}.jinja
@@ -30,6 +30,11 @@ requires-python = "{{ python_version }}"
 dependencies = [
     "snakemake >= {{ snakemake_version }},<8",
     "snakebids >= {{ snakebids_version }}",
+    {#- newer pulps are incompatible with old snakemakes, and we need to support
+    old snakemakes for python versions <3.11. So cap pulp to the last working
+    version
+    #}
+    "pulp < 2.8.0",
 ]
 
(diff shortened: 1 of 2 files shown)


kaitj (approved):
PR lgtm. 

I'd consider adding an early deprecation message (towards the EOL of py3.9) if the plan is to drop support for py3.10 and move to supporting py3.11. Or can go the NumPy route - I think they always support the last supported version +1 (e.g. if Python is 3.8, they support 3.9+). Anyways, a discussion to be had later on.

pvandyken:
> I'd consider adding an early deprecation message (towards the EOL of py3.9) if the plan is to drop support for py3.10 and move to supporting py3.11. Or can go the NumPy route - I think they always support the last supported version +1 (e.g. if Python is 3.8, they support 3.9+). Anyways, a discussion to be had later on.

My idea was not necessarily to fully drop 3.9 early, just in the template app. But that's a few years from now anyway, so no need to make a final decision now

PR #368, opened 2024-02-03
Exclude `run` from test strategy

pvandyken:
This fixes the failing test on main.

The pybids internal formatting of values as integers caused problems

### DIFF
--- a/snakebids/tests/test_generate_inputs.py
+++ b/snakebids/tests/test_generate_inputs.py
@@ -710,8 +710,9 @@ class TestFilterMethods:
             name="template",
             min_entities=2,
             max_entities=2,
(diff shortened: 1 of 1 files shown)


kaitj (approved):
This looks good to me.

I guess something to note if we want to test the `run` entity (we might already be elsewhere), is that the bids spec expects this value to be a non-negative integer (hence the `pybids` expectation).

pvandyken:
> I guess something to note if we want to test the run entity (we might already be elsewhere), is that the bids spec expects this value to be a non-negative integer (hence the pybids expectation).

The issue is more that pybids converts it into an integer internally, rather than simply validating and storing as a string. This prevents us from using it in any test requiring interconversion between path and layout.

PR #376, opened 2024-02-19
Allow custom snakebids version in app template

pvandyken:
Adds flag `--snakebids-version` to the `snakebids create` cli. It can be set to either a version specifier (e.g. `>=a.b`, `==x.y.z`) or a url (e.g. `@ scheme://domain-or-path/....`). Paths and git repositories are identified and properly formatted for a poetry `pyproject.toml`. It does not handle markers (following a `;`) or extras (in between `[...]`). Some minimal validation in the CLI command will help catch these.

This is primarily intended for testing: with this improvement we can specify the path the snakebids repo in our template dry run tests, allowing us to test the current repository with all installation methods. Previously it was downloading snakebids from pip, but this causes errors in e.g. #362, where the template itself is updated and no longer compatible with the pypi installation.

Nevertheless, the cli command is documented and tested and suitable for public use within its defined scope.

(Note: technically this is an enhancement, but I'm labeling maintenance since the primary focus here is internal)


kaitj (approved):
Given the `snakebids` ver

PR #377, opened 2024-02-19
Fix warning messages

pvandyken:
All the warnings were displaying the wrong stack level: now corrected to show user code rather than snakebids code.

The warnings relating to bids were poorly worded or slightly inaccurate. These have been corrected

### DIFF
--- a/snakebids/paths/_factory.py
+++ b/snakebids/paths/_factory.py
@@ -70,11 +70,11 @@ def _handle_subses_dir(
             "will be removed in a future release. Builder functions without "
             "directories can be created using the bids_factory and spec "
             "functions:\n"
-            "   from snakebids.paths import set_bids_spec, specs\n"
-            "   set_bids_spec(specs.v0_0_0(subject_dir=False, "
-            "session_dir=False))"
-        )
-        warnings.warn(wrn_msg, stacklevel=1)
+            "\tfrom snakebids.paths import bids_factory, specs\n"
+            "\tbids_ses = bids_factory(specs.v0_0_0(subject_dir=False))\n"
+            "\tbids_ses(...)\n"
+        ).expandtabs(4)
+        warnings.warn(wrn_msg, stacklevel=4)
     if newspec:
         return bids_factory(newspec)(
             root,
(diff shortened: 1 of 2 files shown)


snakebids/paths/_factory.py
```diff
@@ -246,6 +234,21 @@ def bids(
 
         tail = f"_{suffix}{extension or ''}" if suffix else extension or ""
 
-        return os.path.join(*path_parts) + tail
+        result = os.path.join(*path_parts) + tail
+        if custom_parts and _implicit and not in_interactive_session():
+            wrn_msg = (
+                f"Path generated with custom entities not part of the default BIDS "
```
kaitj, line 240: The wording of this warning message is a little confusing to me. A few questions below that can hopefully clear some of this up:

* What does a "default BIDS spec" mean here (latest)?
* When talking about how "a spec has not been explictly declared", should this mention that we mean a snakebids' spec?
* Related to the previous comment, clarification of "default spec" and "future specs" though I think these may have more context given the mention of upgrading snakebids to start the sentence.

kaitj (commented):
Largely looks good to me - just wanted to clarify some wording in the warning message. It can get a little bit confusing when talking about both BIDS spec and snakebids` spec in the same message.

snakebids/paths/_factory.py
pvandyken:
```py
            wrn_msg = (
                f"Path generated with unrecognized entities, and a snakebids spec has"
                "not been explicitly declared. This could break in future snakebids "
                "versions, as the default spec can be changed without warning.\n"
                f"\tpath = {result!r}\n"
                f"\tentities = {custom_parts!r}\n\n"
                "Please declare a spec using:\n"
                "\tfrom snakebids import set_bids_spec\n"
                f'\tset_bids_spec("{specs.LATEST}")\n'
            ).expandtabs(4)
```

Less wordy. Hopefully "snakebids spec" is precise enough without needing a full definition, I think it's better to keep warning text brief. I could link the docs, but perhaps the code sample is enough?

pvandyken:
Also just fixed the error message for when multiple paths are found. That one's been bothering me for a while

snakebids/paths/_factory.py
kaitj: I think this looks good to me. I tink the code sample is probably enough here, we could always look into linking the docs if necessary if users raise an issue about it.

PR #378, opened 2024-02-19
Expand paths unaltered when comp has no wcards

### LINKED ISSUE #371: `BidsComponent.expand()` returns nothing when no ziplist data
This case corresponds to a path with no wildcards. `expand()` should instead return the path as is.

pvandyken:
Previously returned an empty list. This is different from a component with no entries (but with wildcards), which should return an empty list

Resolves #371


pvandyken:
Test failure waiting on #377

kaitj (approved):
Lgtm

PR #382, opened 2024-02-20
Prevent custom entities from drawing standard ent

pvandyken:
A failing test drew "ce" by chance, which is a valid entity name


kaitj (approved):
This lgtm - almost missed the added `not in (... | set(std_entities))` in `custom_ents`.

pvandyken:
> This lgtm - almost missed the added not in (... | set(std_entities)) in custom_ents.

Ah yes, sorry, I thought it good to refactor that section

PR #383, opened 2024-02-21
Add dependabot config

kaitj:
## Proposed changes

Noted there were still some workflow using old versions (e.g. `actions/checkout@v3`). Setup dependabot config for automatically updating github actions versions and added a `skip_changelog` label to it in order to leave update of github actions out of the release notes. This should make managing the actions versions a little bit easier given the number of workflows we have.

We can consider adding adding the ability to update poetry dependencies to this as well, but I'm not entirely sure how that works. I think it may constrain it to the versions specified by the `pyproject.toml`, but can look into this more in a future PR if we want to go down this road.

## Checklist

## Notes

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,18 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    # Workflow files stored in the default location of `.github/workflows`. 
+    # (You don't need to specify `/.github/workflows` for `directory`. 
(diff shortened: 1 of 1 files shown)


pvandyken:
I do periodically check the poetry deps, there's a simple command to do so, so I don't see that as a big deal.

This is nice though, I had started to try this before and got intimidated

pvandyken (approved):

PR #390, opened 2024-02-23
Update the warning format to be more readable

pvandyken:
I find the default formatting of warnings by the python module a bit unreadable, especially given our current tendency to do multiline, complex explanations (for instance, it prints the problematic line after the warning message, which is confusing).

The warnings module does allow reassignment of the format function, and that's what I've attempted here. A screenshot is attached below. There is a bit of a problem with this though, because any other code throwing warnings would be affected, as there's only a single warning module. @kaitj, if you're not comfortable with this, we can close this, this isn't a huge deal.

image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -175,7 +175,7 @@ omit = ["snakebids/project_template/**", "snakebids/tests/**"]
 
 [tool.coverage.report]
 exclude_lines = [
-    "pragma: not covered",
+    "pragma: no cover",
     "@overload",
     'class [a-zA-Z0-9_]+\([^)]*Protocol.*\)',
     'if TYPE_CHECKING',
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -174,8 +174,7 @@ markers = [
 omit = ["snakebids/project_template/**", "snakebids/tests/**"]
 
 [tool.coverage.report]
-exclude_lines = [
-    "pragma: no cover",
+exclude_also = [
     "@overload",
     'class [a-zA-Z0-9_]+\([^)]*Protocol.*\)',
     'if TYPE_CHECKING',
--- a/snakebids/__init__.py
+++ b/snakebids/__init__.py
@@ -3,6 +3,8 @@ __version__ = "0.0.0"
 
 __submodules__ = ["core", "paths"]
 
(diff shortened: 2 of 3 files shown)


kaitj:
I'm not opposed to this, it makes it more legible. I'm trying to picture what this might actually look like for other code and I'm having a hard time picturing it aside from being different than what users may be use to.

pvandyken:
If I did this correctly, it really should just change the order of the pieces, plus the (now) yellow color formatting. I implemented it essentially the same as the python implementation with a different format string

pvandyken:
Here's an example with a numpy error:
image

pvandyken:
I'd also like to wait until the test suite is done to see how it affects warnings raised there

kaitj:
Yeah, I kind of quite like this - just takes some getting use to if one if use to the default formatting. Would this also affect the warnings thrown by other imported libraries when using snakebids (is this what you meant by affecting other code)? 

I can see it possibly being an issue if tables are involved, but I don't think that is too much of an issue we need to worry about. I wonder if there is a way we can "turn off" this style of formatting, but I also hesitate to add in more flags / options in the config.

pvandyken:
> Would this also affect the warnings thrown by other imported libraries when using snakebids (is this what you meant by affecting other code)?

It would

pvandyken:
> I wonder if there is a way we can "turn off" this style of formatting, but I also hesitate to add in more flags / options in the config.

If we were concerned enough to think disabling should be an option, then I would say don't it at all.

pvandyken:
> My only concern was with tables being part of the message and the 80 line limit (I think this was actually in an older commit or maybe I am misremembering already?). I don't think that is really an occurence that will happen too often though, if at all.

Right, I see. I'm fine removing the text wrapping too, that's not integral to the change. It's just nice because it respects word boundaries, unlike autowrapping in a terminal

kaitj:
Lets go ahead and make this change then without the text wrapping. I'll wait until the tests are done before doing a proper review + accept the PR.

pvandyken:
Okay, it seems to have no effect on warnings thrown during pytest. I guess pytest must use it's own formatter or something. But this is good I think, it's devs working interactively who really need to see the warnings

kaitj (approved):

PR #401, opened 2024-04-09
Updates to get working with snakemake>=8.0

akhanf:
## Proposed changes

I'm using some snakemake v8 features and plugins so wanted to see if we could update to become compatible with it. 
Turns out the required changes seem to be pretty minimal, just a couple things that have been moved around:
- load_configfile() is now in snakemake.common.configfile instead of snakemake.io
- main() is now in snakemake.cli instead of snakemake

I also updated the pyproject since snakemake 8 requires at least python 3.11, and removed any version pinning specific to python versions lower than 3.11. 

NOTE: Still doesn't pass quality checks (pyright stubs), other typing stuff. My local tests were also intermittently failing, but seemed to be going over the runtime.  @pvandyken do you mind having a look at this?

## Checklist


pvandyken:
Thanks for going over this. I was waiting until #362 was merged as I think it overlaps with this and I wanted to save doing it twice. But I haven't had time... on my list of things to do next week when back from vacation. Can this wait till then (just use a local branch in the meantime)?

pvandyken:
I also don't want to pin to 3.11 yet. There's a few issues that need to be worked around but I've more or less worked out a plan that will let us keep our version support intact. Obviously have to try it yet

akhanf:
Yep, no worries! I also figured this might be best merged after the major update.. Yes would be great if we didn't have to pin to 3.11!

pvandyken:
Closed for #406

PR #449, opened 2024-07-17
Add codespell support (config, workflow to detect/not fix) and make it fix few typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,25 @@
+# Codespell configuration is within pyproject.toml
+---
+name: Codespell
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+permissions:
+  contents: read
+
+jobs:
+  codespell:
+    name: Check for spelling errors
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Annotate locations with typos
+        uses: codespell-project/codespell-problem-matcher@v1
+      - name: Codespell
+        uses: codespell-project/actions-codespell@v2
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,3 +31,10 @@ repos:
         language: system
         exclude: ^docs
         types_or: [cython, pyi, python]
+  - repo: https://github.com/codespell-project/codespell
+    # Configuration for codespell is in pyproject.toml
+    rev: v2.2.6
+    hooks:
+    - id: codespell
+      additional_dependencies:
+      - tomli
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Snakebids is a Python package that extends [Snakemake](https://snakemake.github.
 
 Snakebids includes all of the features of Snakemake, including flexible configuration, parallel execution, and Docker/Singularity support, plus:
 
-- **Built-in support for BIDS datasets**: Seamless workflow functionality with a wide range of BIDS datasets, accomodating various levels of complexity.
+- **Built-in support for BIDS datasets**: Seamless workflow functionality with a wide range of BIDS datasets, accommodating various levels of complexity.
 - **BIDS App Creation**: Provide command-line invocations of your workflow following BIDS App guidelines, ensuring reproducibility and enhancing accessibility of your workflow.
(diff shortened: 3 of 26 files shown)


typings/ruamel/yaml/emitter.pyi
```diff
@@ -128,7 +128,7 @@ class Emitter:
         indicator: Any,
         need_whitespace: Any,
         whitespace: bool = ...,
-        indention: bool = ...,
+        indentation: bool = ...,
```
yarikoptic, line 131: generated file, so might need to get to the "source" and skip this one?

pvandyken: This file represents 3rd party code, so nothing I can do for this. Anything in typings should actually be ignored

pvandyken:
Thanks, this should be fine. I'll take a look when I get a chance

typings/ruamel/yaml/emitter.pyi
yarikoptic: ok, added `typings` to be skipped, redid `codespell`, forcepushed!

pvandyken (approved):
LGTM, thanks so much for setting this up

pvandyken:
Looks like there's some linting errors (line length). Seems to be in docstrings, so I don't think ruff will fix automatically, they'll need to be reshaped with a line wrap tool.

yarikoptic:
> Looks like there's some linting errors (line length). Seems to be in docstrings, so I don't think ruff will fix automatically, they'll need to be reshaped with a line wrap tool.

not poetry or ruff user, so your contribution to this PR would be most appreciated ;)

```shell
❯ poetry run ruff format snakebids
Skipping virtualenv creation, as specified in config file.
warning: `ruff format` is a work-in-progress, subject to change at any time, and intended only for experimentation.
ruff failed
  Cause: TOML parse error at line 215, column 12
    |
215 | [tool.ruff.lint]
    |            ^^^^
unknown field `lint`, expected one of `allowed-confusables`, `builtins`, `cache-dir`, `dummy-variable-rgx`, `exclude`, `extend`, `extend-exclude`, `extend-include`, `extend-ignore`, `extend-select`, `extend-fixable`, `extend-unfixable`, `external`, `fix`, `fix-only`, `fixable`, `output-format`, `force-exclude`, `ignore`, `ignore-init-module-imports`, `include`, `line-length`, `tab-size`, `logger-objects`, `required-version`, `respect-gitignore`, `select`, `show-source`, `show-fixes`, `src`, `namespace-packages`, `target-version`, `preview`, `task-tags`, `typing-modules`, `unfixable`, `flake8-annotations`, `flake8-bandit`, `flake8-bugbear`, `flake8-builtins`, `flake8-comprehensions`, `flake8-copyright`, `flake8-errmsg`, `flake8-quotes`, `flake8-self`, `flake8-tidy-imports`, `flake8-type-checking`, `flake8-gettext`, `flake8-implicit-str-concat`, `flake8-import-conventions`, `flake8-pytest-style`, `flake8-unused-arguments`, `isort`, `mccabe`, `pep8-naming`, `pycodestyle`, `pydocstyle`, `pyflakes`, `pylint`, `pyupgrade`, `format`, `per-file-ignores`, `extend-per-file-ignores`
```
or give me a hand here how to tame the poetry etc.

pvandyken:
> not poetry or ruff user, so your contribution to this PR would be most appreciated ;)

Of course! Didn't want to touch your branch without your permission, but I'm happy to finish it off from here!

yarikoptic:
FWIW for above -- I have ancient ruff system wide... installed recent one in venv, but `ruff check snakebids scripts --fix` just checks and does not fix, expected? (I see you pushed, thanks! just want to learn howto for future)

pvandyken:
>  just checks and does not fix, expected? (I see you pushed, thanks! just want to learn howto for future)

Yeah, ruff won't touch docstrings or comments, so I just use a random line re-wrapping extension in my IDE.

pvandyken:
Also the checks will probably fail because of some issue that came up in the nightly. Still have to look into this. But I know this PR doesn't change any code, so I'll just merge either way

Issue #451, opened 2024-07-17
Support for Cloud Storage URLs (with universal_pathlib)


pvandyken:
### Discussed in https://github.com/khanlab/snakebids/discussions/450

<div type='discussions-op-text'>

<sup>Originally posted by **@Karl5766** July 17, 2024</sup>
Ali and I are considering using Google Cloud buckets (with a access url like "gs://path/to/file") to read and write OME_ZARR files in snakemake workflows. We would like something as follows:
```
from snakebids.core.input_generation import generate_inputs
inputs = generate_inputs(
    'gs://khanlab-lightsheet/data/onuska_lifecanvas/bids',  # instead of a local path, we pass in a string starting with "gs://"
    {'spim': {'filters': {'suffix': 'SPIM', 'subject': 'o21', 'extension': 'ome.zarr'}, 'wildcards': ['subject', 'sample', 'acquisition', 'staining']}}
)
print(inputs)
```
while both pybids and this repo are currently depending on Python's Path class, Ali's idea is to use [universal_pathlib's UPath](https://github.com/fsspec/universal_pathlib) objects as drop-in replacement for Path. For pybids, We made a [fork](https://github.com/bids-standard/pybids/compare/master...Karl5766:pybids:master) of pybids which supports UPath and passes the unit tests of the main repository. 

For this repo, after further modify the `os.path.isabs(path)` call in [input_generation.py](https://github.com/khanlab/snakebids/blob/main/snakebids/core/input_generation.py) to make sure `gs://...` is not recognized as a local relative path, the above code seems to work.

Depending on a pybids' fork is going to break dependency for other users of this repo (We are considering a PR to the main branch of pybids for UPath support. Right now, however, it's a fork and it's not finalized), but I think a commit replacing os.path.isabs() with an equivalent expression compatible with cloud storage paths is possible (so we can use a UPath compatible version of pybids in our own project if we need UPath support in snakebids). @pvandyken What do you think?

Edit: I modified [input_generation.py](https://github.com/khanlab/snakebids/blob/main/snakebids/core/input_generation.py)'s line 726 to use `path.absolute() != path` to test for relative path which is a workaround but I think there is probably a better way to do it.</div>

<div type="discussions-op-text">
<sup>Response from @pvandyken July 17, 2024</sup>
UPath is a subclass of path, so should be sufficient to change that line to `path.is_absolute()` (after making sure the path is a pathlib). I would coerce to path immediately in the `generate_inputs()` function. 
[...]

PR #452, opened 2024-07-18
Add compatibility for urls in `generate_inputs`

Karl5766:
## Proposed changes

Related to issue #451, UPath object does not support os.path.is_abs() or .is_absolute() interfaces due to relative vs. absolute paths are not conceptually defined for network locations. This commit changes test for relative path by making sure UPath objects are not tested on whether they are relative path or not.

## Checklist

- [ √ ] Changes have been tested to ensure that fix is effective or that a feature works.
- [ √ ] I have included necessary documentation or comments (as necessary)
- [ √ ] Any dependent changes have been merged and published

### DIFF
--- a/snakebids/core/input_generation.py
+++ b/snakebids/core/input_generation.py
@@ -392,6 +392,24 @@ def _all_custom_paths(config: InputsConfig):
     return all(comp.get("custom_path") for comp in config.values())
 
 
+def _is_local_relative(path: Path | str):
+    """Test if a path location is local path relative to the current working directory.
+
+    Parameter
+    ---------
+        path
+            A UPath, Path, or str object to be checked
+
+    Returns
+    -------
+    is_url : bool
+        True if the path is relative
+    """
+    path_str = str(path)
+    is_doubleslash_schemed = "://" in path_str
+    return not is_doubleslash_schemed and not os.path.isabs(path_str)
+
+
 def _gen_bids_layout(
     *,
     bids_dir: Path | str,
@@ -723,7 +741,7 @@ def _parse_bids_path(path: str, entities: Iterable[str]) -> tuple[str, dict[str,
     # If path is relative, we need to get a slash in front of it to ensure parsing works
     # correctly. So prepend "./" or ".\" and run function again, then strip before
     # returning
-    if not os.path.isabs(path) and get_first_dir(path) != ".":
+    if _is_local_relative(path) and get_first_dir(path) != ".":
         path_, wildcard_values = _parse_bids_path(os.path.join(".", path), entities)
         return str(Path(path_)), wildcard_values
 
--- a/snakebids/tests/strategies.py
+++ b/snakebids/tests/strategies.py
@@ -37,10 +37,22 @@ alphanum = ascii_letters + digits
 valid_entities: tuple[str, ...] = tuple(BidsConfig.load("bids").entities.keys())
 
 path_characters = st.characters(blacklist_characters=["/", "\x00"], codec="UTF-8")
+# StackOverflow answer by Gumbo
+# https://stackoverflow.com/questions/1547899/which-characters-make-a-url-invalid#1547940
+scheme_characters = st.sampled_from(
+    "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-._~?#[]@!$&'()*+,;="
+)
 
 
 def nothing() -> Any:
     return st.nothing()  # type: ignore
+
+
+def schemes() -> st.SearchStrategy[str]:
+    # Generate the prefix part of a url.
+    random_st = st.text(scheme_characters, min_size=1).map(lambda s: f"{s}://")
+    fixed_st = st.sampled_from(["C://", "c:\\", "gs://", "s3://"])
+    return random_st | fixed_st
 
 
 def paths(
@@ -71,6 +83,30 @@ def paths(
         return result.map(lambda p: p.resolve())
 
     return result
+
+
+def schemed_paths(
+    *,
+    min_segments: int = 0,
+    max_segments: int | None = None,
+) -> st.SearchStrategy[str]:
+    """schemed paths are those beginning with scheme://.
+
+    These paths are incompatible with paths() because pathlib.Path turns // into /
+    when parsing, and paths() returns Path objects by default. That's why we have a
+    separate function for schemed paths.
+
+    This function returns un-schemed paths as well, occasionally
+    """
+    scheme_st: st.SearchStrategy[str] = (st.none() | schemes()).map(
+        lambda s: "" if s is None else s
+    )
+    paths_st: st.SearchStrategy[str] = paths(
+        min_segments=min_segments,
+        max_segments=max_segments,
+        absolute=False,
+    ).map(lambda p: str(p))
+    return st.tuples(scheme_st, paths_st).map(lambda t: t[0] + t[1])
 
 
 def bids_entity(
--- a/snakebids/tests/test_generate_inputs.py
+++ b/snakebids/tests/test_generate_inputs.py
@@ -12,7 +12,7 @@ import sys
 import tempfile
 import warnings
 from collections import defaultdict
-from pathlib import Path
+from pathlib import Path, PosixPath
 from typing import Any, Iterable, Literal, NamedTuple, TypedDict, TypeVar, cast
 
 import attrs
@@ -30,6 +30,7 @@ from snakebids.core.input_generation import (
     _all_custom_paths,
     _gen_bids_layout,
     _get_components,
+    _is_local_relative,
     _normalize_database_args,
     _parse_bids_path,
     _parse_custom_path,
@@ -1602,6 +1603,49 @@ def test_all_custom_paths(count: int):
         assert not _all_custom_paths(config)
 
 
+class TestRecogPathSchemes:
+    PATH_AND_TYPES = (
+        ("file", "RELATIVE"),
+        ("hello", "RELATIVE"),
+        ("gs", "RELATIVE"),
+        ("./hello/world", "RELATIVE"),
+        ("hello/world", "RELATIVE"),
+        ("/hello/world", "ABSOLUTE"),
+        ("gs://some/google/cloud/bucket", "NETWORK"),
+        ("s3://some/s3/bucket", "NETWORK"),
+    )
+
+    @pytest.mark.parametrize(("path", "path_type"), PATH_AND_TYPES)
+    def test_is_local_relative(self, path: str, path_type: str):
+        isnet = path_type == "NETWORK"
+        is_local_relative = path_type == "RELATIVE"
+
+        # test the path itself, and the corresponding Path(path)
+        assert (
+            _is_local_relative(path) == is_local_relative
+        ), f"Path {path} fails is local relative path test."
+        if not isnet:
+            assert _is_local_relative(Path(path)) == is_local_relative
+
+    @pytest.mark.skipif(
+        sys.version_info < (3, 12), reason="Path class has no with_segments()"
+    )
+    @pytest.mark.parametrize(
+        ("path", "path_type"), [tup for tup in PATH_AND_TYPES if tup[1] == "RELATIVE"]
+    )
+    def test_path_subclassing(self, path: str, path_type: str):
+        # Google cloud is not posix, for mocking purpose however we just
+        # need a class that is a subclass of Path
+        class MockGCSPath(PosixPath):
+            def __init__(self, *pathsegments: str):
+                super().__init__(*pathsegments)
+
+            def __str__(self):  # __fspath__ calls __str__ by default
+                return f"gs://{super().__str__()}"
+
+        assert not _is_local_relative(MockGCSPath(path))
+
+
 @example_if(
     sys.version_info >= (3, 8),
     dataset=BidsDataset(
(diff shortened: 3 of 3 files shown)


snakebids/core/input_generation.py
```diff
@@ -392,6 +393,30 @@ def _all_custom_paths(config: InputsConfig):
...
+    path
+        A UPath, Path, or str object to be checked
+
+    Returns
+    -------
+    is_url : bool
+        True if the path is a url
+    """
+    path = str(path)
+    if path.startswith(("file://", "local://")):
```
pvandyken, line 413: When these get read in by the `_is_url(...) and not Path(...)`, they'll break (they won't appear to be absolute).

The absolute check below is really just to make sure that `sub-XXX` is not the very first part of the path, because we need a `/` character before `sub`. This isn't an issue with urls.

I would say just treat anything with a `scheme://` at the beginning as a URL, regardless of whether the scheme is `file`.

snakebids/tests/test_generate_inputs.py
```diff
@@ -1602,6 +1603,35 @@ def test_all_custom_paths(count: int):
...
+    for path, path_type in test_data:
+        # https://stackoverflow.com/questions/8357098/how-can-i-check-if-a-url-is-absolute-using-python
+        isurl = path_type == "URL"
+
+        # test the path itself, and the corresponding Path(path)
+        assert (
+            _is_url(path) == isurl
+        ), f"Path {path} should have isurl={isurl} but got {_is_url(path)}"
+        if not isurl:
+            assert not _is_url(Path(path))
```
pvandyken, line 1632: If you're willing to have a go at hypothesis, could you expand the `TestParseBidsPath` tests to potentially include schemes?

As a hint, basically I would add a new parameter `scheme` with a text strategy filling it with any [valid URI character] (https://stackoverflow.com/questions/1547899/which-characters-make-a-url-invalid#1547940) except for `:/\`, or with `None` (something like `st.text(...) | st.none()`). If `None`, don't prepend a scheme, otherwise prepend the scheme with a valid separator. The test should then work the same either way.

snakebids/core/input_generation.py
pvandyken: With this change, we don't actually need this function. Just inline the relevant bit and expand the `TestParseBidsPath` class as discussed below.

pvandyken (changes requested):
Other than the comments, this is good, thanks!

pvandyken:
The test failures were popping up on the nightly, they're nothing to do with this PR. I'll have to investigate when I get a chance, don't worry about fixing them in this PR

snakebids/core/input_generation.py
Karl5766: `_is_url(...) and not Path(...)` tests if the path is a relative path, but looking at [wiki](https://en.wikipedia.org/wiki/File_URI_scheme) I think "file:///path" can be both relative or absolute depending on what follows the "file:///" prefix. I'm keeping the _is_url() function as a interface (non-duplicated definition of what's a url).

If we treat `scheme://` as URL, then windows path `C://` or Posix `c:\\` will be treated as URL, which defeats the purpose. Are "file:///" and "local://" paths currently supported? If not, maybe we treat them as unsupported types until they are needed in the future. With this I will remove the above two lines testing specifically for them.

snakebids/tests/test_generate_inputs.py
Karl5766: I see the synthetic bids path functions currently accept a `root: Path | None = None` parameter that defaults to the current `"."` path, but how do I combine it with paths generated from a `path()` strategy?

Karl5766:
Just updated the PR for the changes in above 2 comments. There may be issues as mentioned so let me know what you think.

snakebids/tests/test_generate_inputs.py
pvandyken: Ah, I see, I wouldn't worry about the root. What we care about testing is the scheme `scheme://`, because that's the feature that's not currently possible that your pull request enables. So from my perspective, I wouldn't make a new composite strategy, but something like this:

```py
    @given(
		component=sb_st.bids_components(max_values=1, restrict_patterns=True),
		scheme=st.text(st.characters(...)) | st.none(), # just use valid scheme characters here
	)
    def test_splits_wildcards_from_path(self, component: BidsComponent, scheme: str | None):
        path = component.expand()[0]
		if scheme is not None:
			path = f"{scheme}://{path}"
        ...
```

I think something like that should do the trick.

pvandyken: I guess technically we could expand the `bid_components()` strategy to potentially produced paths with schemes, but since it's really just this one function affected, I don't see a ton of utility right now for the extra complexity it creates

Karl5766: Just pushed a commit to fix this

snakebids/tests/strategies.py
```diff
@@ -43,12 +43,27 @@ def nothing() -> Any:
     return st.nothing()  # type: ignore
 
 
+def schemes() -> st.SearchStrategy[str | None]:
```
pvandyken, line 46: It's best to have a strategy return one type of thing. So have this just return strings, then when you use it, you would mix it with a `None` strategy

snakebids/tests/strategies.py
```diff
@@ -43,12 +43,27 @@ def nothing() -> Any:
...
+    return st.one_of(st.none(), fixed)
+
+
 def paths(
     *,
     min_segments: int = 0,
     max_segments: int | None = None,
     absolute: bool | None = None,
     resolve: bool = False,
+    prefix_st: st.SearchStrategy[str | None] = None,  # for urls e.g. "gs://..."
```
pvandyken, line 66: Can we instead just treat schemes the way we treat absolute? So have a `scheme: bool | None` parameter, defaults to `None`. If `None`, it may or may not add a scheme. If false, it doesn't, if true, it does. This will help make sure schemes are tested wherever we use paths.

snakebids/tests/strategies.py
```diff
@@ -60,6 +75,14 @@ def paths(
 
     absolute_paths = paths.map(lambda p: Path("/", p))
 
+    # append scheme prefix if argument prefix_st is provided
+    if prefix_st:
```
pvandyken, line 79: This builds on the above comment, but for the interaction with `absolute`, only complain if the user specifies incompatible options like `scheme=True`, `absolute=True`. If one or both are `None`, just implicitly set the other as needed

snakebids/core/input_generation.py
```diff
@@ -723,7 +746,11 @@ def _parse_bids_path(path: str, entities: Iterable[str]) -> tuple[str, dict[str,
     # If path is relative, we need to get a slash in front of it to ensure parsing works
     # correctly. So prepend "./" or ".\" and run function again, then strip before
     # returning
-    if not os.path.isabs(path) and get_first_dir(path) != ".":
+    if (
+        not _is_network_fs_path(path)
```
pvandyken, line 750: So I still don't understand why we care about network paths specifically in this function. The function isn't actually reading the url, it's parsing it as a string. And the only reason we had the relative path check before is that we can't have `sub-whatever` as the very first thing in the path, because the regex expects a slash first `/sub-whatever`. So so long as we start with any kind of scheme (even `c//` or `C:\\`), we should be fine?

snakebids/core/input_generation.py
```diff
@@ -438,7 +461,7 @@ def _gen_bids_layout(
     if not pybidsdb_dir:
         pybidsdb_dir = None
     # Otherwise check for relative path and update
-    elif not Path(pybidsdb_dir).is_absolute():
+    elif not _is_network_fs_path(pybidsdb_dir) and not Path(pybidsdb_dir).is_absolute():
```
pvandyken, line 464: Will pybids be able to read network paths for the **database**, even with your PR? If so this is fine, but otherwise I don't care to build support for something that doesn't work in pybids anyway.

snakebids/tests/strategies.py
```diff
@@ -43,12 +43,27 @@ def nothing() -> Any:
...
+    # Generate the prefix part of a url.
+    #
+    # we only sample from a fixed set of schemes. If schemes are randomly generated
+    # then something like file:// could be a problem because it may be either an
+    # abs or relative path and will fail the os.path.isabs() code
+    #
+    # Note for its current status we don't guarantee the following schemes to work
+    # but only requiring them to be retained correctly i.e. not from 'gs://' to
+    # 'gs:/' when fed into generate_inputs() and path expansion
+    fixed = st.sampled_from(["C://", "c:\\", "gs://", "s3://"])
```
pvandyken, line 56: Can you make the letters (`gs`, `s3` etc) random? So draw something from the `text` strategy.

snakebids/tests/test_generate_inputs.py
```diff
@@ -1602,6 +1603,46 @@ def test_all_custom_paths(count: int):
...
+        ("hello/world", "RELATIVE"),
+        ("/hello/world", "ABSOLUTE"),
+        (r"C:\some\file", "ABSOLUTE"),
+        (r"C:\\some\\file", "ABSOLUTE"),
+        ("C:/some/file", "ABSOLUTE"),
+        ("gs://some/google/cloud/bucket", "NETWORK"),
+        ("s3://some/s3/bucket", "NETWORK"),
+    )
+
+    if sys.version_info[:2] >= (3, 12):
```
pvandyken, line 1621: It's best to split this off into a separate test, and use `pytest.mark.skipif` (there's some examples in this codebase, and also easy to find online).

You needn't test every example with the subclass, one will suffice.

snakebids/tests/test_generate_inputs.py
```diff
@@ -1602,6 +1603,46 @@ def test_all_custom_paths(count: int):
         assert not _all_custom_paths(config)
 
 
+def test_is_network_fs_path():
+    test_data = (
```
pvandyken, line 1607: My preference is to make tests as small and granular as possible. With that in mind, these examples should be run using `@pytest.mark.parametrize`.

snakebids/tests/test_generate_inputs.py
```diff
@@ -1865,17 +1906,38 @@ def test_when_all_custom_paths_no_layout_indexed(
     spy.assert_not_called()
 
 
+@st.composite
+def scheme_bids_component_strategy(draw, **kwargs):
```
pvandyken, line 1910: Make sure this gets removed if its not being used

pvandyken (changes requested):
Left another round of comments, but it's coming together!

snakebids/core/input_generation.py
pvandyken: Just left another comment related, but to close this off, I think the confusion might be coming because snakebids generally doesn't care about schemes at all. What this PR is addressing is a pretty small technicality relating to how we parsed paths before. We really just need to check that the very first character in the path is not part of an entity. But if we have `anything://` at the beginning, we know for sure that's not an entity. So I was questioning why `file://` was getting treated specially.

Windows support is a bit of another fish. I don't mind adding code to support drive letters, but until we actually run tests on a windows image on github, we don't promise any windows support. Not really a priority right now because almost all apps realistically are on either linux, WSL, or some docker image.

snakebids/core/input_generation.py
Karl5766: For this one, "gs://" will be converted to "gs:/" that's not supported by gcsfs, that's really the only issue here

snakebids/core/input_generation.py
Karl5766: Can you point me to docs about paths with databases? I'm not particularly well versed in this topic.

Karl5766: Ah I see what you mean, I'll just remove this line

snakebids/tests/strategies.py
Karl5766: Mostly I want to avoid issues that some file schemes behave differently than others (`file` can be `file:/`, `file://` or `file:///` and may be either local or mapped from network), so only test on a fixed set of schemes. I guess this PR is really asking for a `is_local_relative` function that tests if a path is a relative path to the current working directory. Should I change `is_network_fs_path` to this `is_local_relative` function?
```
def _is_network_fs_path(path: Path | str):
    return not bool(urlparse(str(path)).netloc) and not os.path.isabs(path)
```

snakebids/tests/test_generate_inputs.py
```diff
@@ -1866,16 +1912,30 @@ def test_when_all_custom_paths_no_layout_indexed(
 
 
 class TestParseBidsPath:
-    @given(component=sb_st.bids_components(max_values=1, restrict_patterns=True))
-    def test_splits_wildcards_from_path(self, component: BidsComponent):
+    @given(
+        component=sb_st.bids_components(max_values=1, restrict_patterns=True),
+        scheme=sb_st.schemes(),
```
pvandyken, line 1917: I think this and the others need the `st.none()` strategy as well, otherwise

snakebids/core/input_generation.py
pvandyken: So that's what I'm concerned about. Right now (as per my tests), `file:///...` be converted into `file:/...`? That's also undesirable, no? Whereas `file://...` will be preserved.

pvandyken (changes requested):

snakebids/tests/strategies.py
pvandyken: Yeah, so I think testing of Windows Paths works best on a windows system. In that case, Python automatically uses the `WindowsPath`, which handles drives correctly. As I mentioned previously, I'm okay with testing windows drive here, but until we actually run the test suite on a windows system, we don't officially support windows

pvandyken:
I think there's still some confusion in the comments, so I'm going to try to lay everything here again as best as I can to make sure we're both talking about the same thing.

The problem in this PR is that certain paths are getting mangled by the `generate_inputs` function, particularly any path with a network scheme (e.g. `scheme://`). This is because the `_parse_bids_path` uses `pathlib.Path` to do some normalization, and these removes any doubled or tripled slashes. This normalization is done to ensure `sub-XXX` is not the very first thing in the path.

So to fix this, we add a function that detects if such a scheme is being used, and if so, we don't do any path normalization. This leaves the path intact. In general, any path with double or more slashes should not be modified. The semantic meaning of the slashes or the scheme is unimportant, we just don't want to change it. The path will then be passed transparently to the `BidsComponent`, where the user can apply it appropriately.

Does this seem like an accurate summary?

Karl5766:
Yes this does summarizes and clarifies things. i've pushed new commits to address these issues, so please take a look if available

pvandyken (approved):
Looks great! We can merge as soon as all the checks are passing. I'm going to get #453 merged first, so you'll be able to rebase onto that

pvandyken:
When fixing the quality checks, do you mind squashing your changes into a single commit with a description of the fix?

pvandyken:
Just merged #453, so you can go ahead and rebase that as well

Karl5766:
> When fixing the quality checks, do you mind squashing your changes into a single commit with a description of the fix?

That is squashing the previous 7 commits if I understand it right? Sure then

pvandyken:
> That is squashing the previous 7 commits if I understand it right? Sure then

The 7 commits.

Karl5766:
I'm leaving a `schemed_paths()` function in strategies.py due to return type issue, but note the function is not called anywhere. @pvandyken Can you merge the PR if everything looks good?

pvandyken:
Thanks. Never got a chance to ask, but what is the purpose of keeping `schemed_paths()` around? I'd rather not keep a function that's never being used.

Karl5766:
Sorry I thought you asked me to add scheme to paths strategy, but upon rereading the comments that's not the case. Yeah you can remove it since it's unused.

PR #459, opened 2024-11-22
Support for conda

Dhananjhay:
## Proposed changes
This PR adds a conditional import for the `ruamel_yaml` dependency. This change is necessary because the conda package manager installs the package as `ruamel_yaml`, while pip installs it as `ruamel`. This discrepancy leads to import errors when working in a conda environment. The conditional import ensures we are agnostic to the installation method, preventing the import error.

Bug Fix:
This PR resolves the import error issue that occurs when the `ruamel.yaml` module is installed via conda. By handling both cases, it ensures the code works in both pip and conda environments.

## Checklist

### DIFF
--- a/snakebids/io/yaml.py
+++ b/snakebids/io/yaml.py
@@ -2,7 +2,10 @@ import collections
 from pathlib import Path, PosixPath, WindowsPath
 from typing import Any, OrderedDict
 
-from ruamel.yaml import YAML, Dumper
+try:
+    from ruamel.yaml import YAML, Dumper
+except ImportError:
+    from ruamel_yaml import YAML, Dumper
 
 
 def get_yaml_io():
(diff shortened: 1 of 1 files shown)


akhanf (approved):
Looks good to me

Dhananjhay:
It'll be nice to close this PR soon and merge the necessary changes now that snakebids v0.13.1 is available on bio-conda ecosystem (https://anaconda.org/bioconda/snakebids)

akhanf:
What would we need to do to get the unit tests passing? Looks like pyright type checking is failing.. I'm not exactly sure why changing how its imported affects this, but if we can sort that out then we can merge in.

Dhananjhay:
@akhanf This new commit skips type checking `ruamel_yaml` package coming from conda - reason why pyright tests were failing because `ruamel_yaml` is `raumel.yaml` in pypi. I was able to run the pyright tests locally without any errors!

akhanf (approved):
Looks good to me, and passing tests, so I will merge this in..

akhanf:
hmm, looks like the tests didn't actually run in the checks (which I thought they did), so they seem to be failing in main now..

PR #466, opened 2025-03-14
Update tutorial.md to reference pulp version issue

jcf2:
Tutorial steps did not work on my Python (3.9) as written, so if continuing to write them for "at least version 3.7" maybe this link to a workaround will help.

## Proposed changes

After getting an error following the tutorial and searching it up, I located a fix in a snakebids issue. Linking to that might be one option to address this error coming up, without interrupting the current nice flow in the doc?

The error is specifically `AttributeError: module 'pulp' has no attribute 'list_solvers'` because `list_solvers` was deprecated in pulp v. 2 and removed in v. 3, and following the tutorial my snakemake install included 3.0.2.

## Checklist

## Notes

### DIFF
--- a/docs/tutorial/tutorial.md
+++ b/docs/tutorial/tutorial.md
@@ -24,7 +24,8 @@ $ mkdir snakebids-tutorial
 $ cd snakebids-tutorial
 ```
 
-Check your python version to make sure you have at least version 3.7 or higher:
+Check your python version to make sure you have at least version 3.7 or higher (extra steps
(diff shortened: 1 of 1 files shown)


jcf2:
Sorry, just realized I listed my local Python incorrectly as "3.8" when it is 3.9 (fixed above).

pvandyken:
Thanks for bringing this up! We've known internally about this for a while but haven't noted it anywhere public yet.

I'll merge this for now. My preference will be to fix dependencies so that pulp install properly (doesn't look like we put an upper cap on it right now...). 

I'm also going to update the tutorial to say minimum 3.9 as that's currently the lowest supported version.

PR #468, opened 2025-04-21
add uv package manager option

Dhananjhay:
## Proposed changes

This PR adds support for the uv package manager as an option when creating a Snakebids app using the `snakebids create command`, alongside existing options like poetry and hatch.

Key changes include:
- Conditional logic in the Jinja template files to configure the pyproject.toml and install instructions for uv
- Adjustments in the docs and templates to ensure compatibility and clarity
- Logic alignment with hatch for backend config (i.e., uses hatchling as the build backend for uv)
- Updated CLI validation logic to recognize uv as a valid option

This resolves support request/feature expansion for modern Python packaging tools and enables compatibility with uv, a fast Python package manager gaining popularity in the community.

## Checklist

## Notes


pvandyken:
@Dhananjhay, are you able to finish this? `test_template_dry_runs_successfully` will need `uv` added as a test case. This will require updating the `test-template` container. You should be able to update the Dockerfile to install uv directly, then use uv to add all the installed tools instead of pipx.

Dhananjhay:
Sounds good! Seems like this branch is out of date as well, will do a quick rebase and make the required changes. Thanks for the feedback!

PR #470, opened 2025-08-15
normalize mixed/absent entities with a configurable “snakenull” label

jordandekraker:
## Proposed changes

Adds optional normalization so that when a component lists an entity in wildcards but some matched files lack that entity, missing values are relabeled to `snakenull` (configurable). If an entity listed in wildcards is entirely absent across the dataset, that entity is skipped for that component. Path construction can include or omit the {key}-snakenull segment via a flag. Default remains the current behavior (feature disabled). This integrates cleanly with `generate_inputs()`/`BidsComponent` (per Snakebids docs), and preserves expand/wildcards semantics users rely on. 

## Checklist

- [x] Changes pass the unit tests (pytest -q snakebids/tests/test_snakenull.py;  poe quality; still working through other quality checks)

## Notes

### DIFF
--- a/docs/bids_app/config_snakenull.md
+++ b/docs/bids_app/config_snakenull.md
@@ -28,17 +28,20 @@ pybids_inputs:
       scope: [session, acquisition]
 ```
 ## Usage
-After calling snakebids.generate_inputs(...), apply:
+include these parameters in generate_inputs():
 ```python
-from snakebids.snakenull import normalize_inputs_with_snakenull
-inputs = snakebids.generate_inputs(bids_dir=..., pybids_inputs=...)
-normalize_inputs_with_snakenull(inputs, config=your_full_config_dict)
+inputs = generate_inputs(
+    bids_dir=config["bids_dir"],
+    pybids_inputs=config["pybids_inputs"],
+    ...
+    snakenull=config.get("snakenull"),
+)
 ```
 This mutates inputs in place so that:
 
-Entities listed in wildcards but entirely absent in your dataset are removed.
+- Entities listed in wildcards but entirely absent in your dataset are removed.
 
-Entities that are present for some files but missing for others are assigned the
+- Entities that are present for some files but missing for others are assigned the
 placeholder label (default snakenull) and included in entities for expansion.
 
 Downstream, your BidsComponent.entities will include e.g.:
--- a/docs/bids_app/config_snakenull.md
+++ b/docs/bids_app/config_snakenull.md
@@ -1,0 +1,51 @@
+# Configuration: Snakenull normalization (optional)
+
+Snakebids can **normalize mixed or absent entities**—cases where some files have an entity
+like `acq` or `ses` but others don’t—by assigning a placeholder label to missing values
+and skipping entities that are entirely absent. This stabilizes wildcard expansion and
+output naming without requiring users to re-label inputs.
+
+This behavior is **off by default** and can be controlled globally or per component.
+
+## Global defaults
+
+```yaml
+snakenull:
+  enabled: false       # default: legacy behavior
+  label: snakenull     # placeholder value for missing entities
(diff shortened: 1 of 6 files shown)


akhanf:
Yes some upstream env issue is not causing the tests to run, can you try running the tests locally? 

The changes are hard to review since the diff is quite messy -- are you able to provide some high-level summary of the implementation-level changes?

jordandekraker:
Maybe you already looked at it now, but here's some further explanation fo the feature from my locally built RTD:
image

This is specifically in response to https://github.com/khanlab/snakebids/issues/469 

Note that I have basically no experience with Typing or pyright. Maybe someone who knows this well and has time can handle this, but otherwise I'll learn. But before I do I just wanted some general feedback. You seemed skeptical in the https://github.com/khanlab/snakebids/issues/469 discussion so i just wanted to make a quick sketchup for you to consider. 

I can try to clean up the messy diffs too, this was mostly an artifact of having chatgpt work on the initial draft.

akhanf:
> Maybe you already looked at it now, but here's some further explanation fo the feature from my locally built RTD:
> 
> This is specifically in response to #469

I read through that and understand what you mean there, it is more a sketch of the implementation that was missing for me. I also don't see how this doesn't break snakebids, since how do you grab bids dataset inputs with expansion BidsComponent.path, now that you don't just have a single unique path pattern? 

I haven't tried running this PR but I imagine the actual tests will be failing, unless you've tried them already? 

> Note that I have basically no experience with Typing or pyright. Maybe someone who knows this well and has time can handle this, but otherwise I'll learn. But before I do I just wanted some general feedback. You seemed skeptical in the #469 discussion so i just wanted to make a quick sketchup for you to consider.
> 
> I can try to clean up the messy diffs too, this was mostly an artifact of having chatgpt work on the initial draft.


Overall I really just want to make sure the code quality in snakebids is maintained --  alot of effort has been put into this in the past,  and proposals for major overhauls like this are typically sketched out in discussion and conceptualized and debated first before just diving in..

jordandekraker:
Ah OK I understand - I'm not used to working with such a high set of development standards :grimacing:

My dryrun testing within micapipe works, generating a clean DAG. 

The logic/sketchup of this feature follows something like this:
1. Discover & parse files
generate_inputs queries PyBIDS and, per matched file, parses only the requested wildcards that exist on that file.
We collect:
  - a set of candidate path templates (with {wildcards}), and
  - a per-file dict of parsed entity values (missing → "").
2. (Conditional) collapse template shapes
If snakenull.enabled is true (globally or per component), we collapse filename shapes by removing optional tag-{tag} tokens that aren’t present for all files (e.g., drop acq-{acq} when only some files have it). If disabled, we keep legacy behavior and raise the classic “Multiple path templates...” ConfigError.
3. Build rectangular zip_lists aligned to the final template
We only include placeholders that actually appear in the chosen template and pad each file’s row so all lists have equal length.
4. Normalize entities (snakenull pass)
If snakenull.enabled is true, we compute entity domains from the per-file dicts:
- Drop entities that are never present across the component (they’re ignored downstream).
- For entities present in some files but missing in others, insert the placeholder label (default snakenull) into that entity’s value set.
- We do not inject snakenull per-file; we add it to the domain so downstream iteration can include it as a value.
- For frozen objects (e.g., BidsComponent), we avoid mutating attrs and stash read-only context instead.

Chatgpt also had these optimistic comments on what we've done here:
Edge cases & guarantees:
- Works with custom_path components (normalizer falls back to zip_lists inspection).
- Collapse only affects filename tokens (keeps directory entities like ses-{session} intact).
- zip_lists are always rectangular and consistent with the final path template.
- Backwards compatible when disabled; tests added for enabled/disabled/mixed cases.

Why this is safe
- We never change which files are included—just how entity domains are exposed when data is heterogeneous.
- When disabled, behavior is identical to prior releases.

PR #477, opened 2025-11-03
Switch to UV and Update Python versions

pvandyken:
Migrate from Poetry to UV for package management, update the minimum Python version from 3.8 to 3.10, and add a new bids spec for Bids 1.10.1.

- Migration from Poetry to UV build system (using hatchling and hatch-vcs).
- Switch to `src/` directory layout. This is better supported by UV, and is really the standard for modern Python projects. Tests now live in a standalone test directory `tests/`.
- Update minimum Python version from 3.8 to 3.10 (dropping 3.8 and 3.9 support). Update legacy code where relevant (not a lot for this update). Add testing for Python 3.13 and 3.14.
- Add strict=True to many zip() calls to explicitly handle length mismatches
- Improve the implementation of `zip_list_eq` for better speed.
- Update all dependencies, especially pybids to 0.21.0.
- Fakefs has been dropped in almost all circumstances because it is not compatible with the `UPath` now used in pybids. Now all tests just run on the filesystem directly. For a future PR, the very chunky dataset tests should really be slimmed down: these are essentially integration tests using hypothesis, very slow, and antipattern. Fakefs can probably be dropped completely.


akhanf:
looks good to me. feel free to drop any PRs that are no longer needed because of this..

akhanf (approved):

PR #481, opened 2025-11-14
Add uv build option for template creation

Dhananjhay:
## Proposed changes

This PR adds support for the uv package manager as an option when creating a Snakebids app using the snakebids create command.

Key changes include:
- Conditional logic in the Jinja template files to configure the pyproject.toml and install instructions for uv
- Logic alignment with uv_build for backend config (i.e., uses uv_build as the build backend for uv)
- Updated CLI validation logic to recognize uv as a valid option

This resolves support request/feature expansion for modern Python packaging tools and enables compatibility with uv. 

## Checklist

## Notes

### DIFF
--- a/src/snakebids/project_template/{% if build_system != 'poetry' %}pyproject.toml{% endif %}.jinja
+++ b/src/snakebids/project_template/{% if build_system != 'poetry' %}pyproject.toml{% endif %}.jinja
@@ -48,6 +48,9 @@ build-backend = "flit_core.buildapi"
 {%- elif build_system == "hatch" -%}
 requires = ["hatchling"]
 build-backend = "hatchling.build"
+{%- elif build_system == "uv" -%}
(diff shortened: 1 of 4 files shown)


src/snakebids/project_template/{% if build_system != 'poetry' %}pyproject.toml{% endif %}.jinja
```diff
@@ -48,6 +48,9 @@ build-backend = "flit_core.buildapi"
 {%- elif build_system == "hatch" -%}
 requires = ["hatchling"]
 build-backend = "hatchling.build"
+{%- elif build_system == "uv" -%}
+requires = ["uv_build>=0.9.9,<0.10.0"]
```
pvandyken: Didn't know about this actually, thanks!

pvandyken (approved):

